PR #1, opened 2020-05-28
Add aarch64 machine slave to build flang with clang-8

CarolineConcatto:
This patch adds another slave to build flang. It uses clang-8 and clang++-8
Signed-off-by: Caroline Concatto <caroline.concatto@arm.com>


CarolineConcatto:
This should be open on llvm.review

PR #4, opened 2020-07-24
Add Vulkan support for mlir-nvidia

ChristianKuehnel:
details:
- updated mlir-nvidia Dockerfile with Vulkan support
- duplicated kubernetes deployment for prod and staging
- fixed deployment script on Linux
- added more log output to run.sh
- fixed host information output
- added cmake flag


ChristianKuehnel:
@antiagainst This updated container seems to be working fine. I was *not* able to manually run the build with `-DMLIR_VULKAN_RUNNER_ENABLED=1` because the container got rebooted repeatedly.

I would say let's just give it a shot: Let's merge it and see if the builds keep passing and revert if anything goes wrong.
What do you think?

antiagainst:
> I would say let's just give it a shot: Let's merge it and see if the builds keep passing and revert if anything goes wrong.
> What do you think?

SGTM! I think it should just work. But yes we can revert if something goes wrong. Thanks so much for the wiring up, @ChristianKuehnel! This has been a long standing issue; great to have this finally addressed!

PR #11, opened 2021-06-16
[libc++] Remove build bots that are handled on BuildKite now

ldionne:
Those builders have been failing for months because they use a version
of Clang that we don't actively support anymore. Furthermore, all those
configurations (single-threaded, no exceptions, with LLVM's unwinder)
are being tested on BuildKite.

At this point, I think those builders are only adding noise and using
resources that could be repurposed to other tasks, so I think it is
reasonable to remove them.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1327,41 +1327,6 @@ all = [
 
 # Libc++ builders.
 
-    {'name': 'libcxx-libcxxabi-x86_64-linux-debian',
-    'tags'  : ["libcxx"],
-    'workernames': ['gribozavr4'],
-    'builddir': 'libcxx-libcxxabi-x86_64-linux-debian',
-    'factory': LibcxxAndAbiBuilder.getLibcxxAndAbiBuilder(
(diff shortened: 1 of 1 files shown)


gribozavr (approved):

maxim-kuvyrkov:
@ldionne , I suggest you post this on https://reviews.llvm.org/ .  llvm-zorg reviews are handled there; see https://llvm.org/docs/Phabricator.html .

ldionne:
I just wanted to have @gribozavr 's approval. Committed as

```
commit 8d8b4bb34168d53a4d0e3eba3614aa0ce9e7fd9e
Author: Louis Dionne <ldionne.2@gmail.com>
Date:   Wed Jun 16 15:42:02 2021 -0400

    [libc++] Remove build bots that are handled on BuildKite now

    Those builders have been failing for months because they use a version
    of Clang that we don't actively support anymore. Furthermore, all those
    configurations (single-threaded, no exceptions, with LLVM's unwinder)
    are being tested on BuildKite.

    At this point, I think those builders are only adding noise and using
    resources that could be repurposed to other tasks, so I think it is
    reasonable to remove them.
```

PR #15, opened 2021-08-12
Replace new-pass-manager-fast builder with llvm-dylib-all

tbaederr:
Test a LLVM build with the dynamic library enabled. Remove the
clang-x86_64-debian-new-pass-manager-fast builder since the new pass
manager has been the default since Feb 21.

I've modeled this after the existing `x86_64-fedora-clang` builder.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -56,30 +56,6 @@ all = [
                         'CC': 'ccache clang', 'CXX': 'ccache clang++', 'CCACHE_CPP2': 'yes',
                     })},
 
-    {'name' : "clang-x86_64-debian-new-pass-manager-fast",
-    'tags'  : ["clang", "fast"],
-    'workernames': ["gribozavr4"],
-    'builddir': "clang-x86_64-debian-new-pass-manager-fast",
-    'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
-                    llvm_srcdir="llvm.src",
-                    obj_dir="llvm.obj",
-                    clean=True,
-                    depends_on_projects=['llvm','clang','clang-tools-extra','compiler-rt'],
-                    extra_configure_args=[
-                        "-DCOMPILER_RT_BUILD_BUILTINS:BOOL=OFF",
-                        "-DCOMPILER_RT_BUILD_ORC:BOOL=OFF",
-                        "-DCOMPILER_RT_BUILD_SANITIZERS:BOOL=OFF",
-                        "-DCOMPILER_RT_BUILD_XRAY:BOOL=OFF",
-                        "-DCOMPILER_RT_INCLUDE_TESTS:BOOL=OFF",
-                        "-DENABLE_EXPERIMENTAL_NEW_PASS_MANAGER=ON",
-                        "-DCMAKE_C_FLAGS=-Wdocumentation -Wno-documentation-deprecated-sync",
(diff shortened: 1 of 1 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -1848,6 +1824,34 @@
                         '-DLLVM_ENABLE_LLD=ON',
                     ])},
 
+# Build the LLVM dylib .so with all backends and link tools to it
+    {'name' : 'llvm-dylib-all',
```
gribozavr, line 1828:
```suggestion
    {'name' : 'llvm-x86_64-debian-dylib',
```

I'm not quite certain what "all" means.

buildbot/osuosl/master/config/builders.py
```diff
@@ -1848,6 +1824,34 @@
...
+                    extra_configure_args=[
+                        '-DCMAKE_BUILD_TYPE=Release',
+                        '-DCMAKE_C_COMPILER=/usr/bin/gcc',
+                        '-DCMAKE_CXX_COMPILER=/usr/bin/g++',
+                        '-DLLVM_ENABLE_ASSERTIONS=On',
+                        '-DLLVM_BUILD_EXAMPLES=Off',
+                        "-DLLVM_LIT_ARGS=-v --xunit-xml-output test-results.xml",
+                        '-DLLVM_CCACHE_BUILD=On',
+                        '-DLLVM_CCACHE_DIR=/ccache',
+                        '-DLLVM_CCACHE_MAXSIZE=20G',
```
gribozavr, line 1846: Please remove the CCACHE flags and DCMAKE_C_COMPILER. Instead, please imitate the configuration of the builder that you're replacing (environment variables).

buildbot/osuosl/master/config/builders.py
tbaederr: I was only building the `x86_64` target previously and changed to build all targets, thus the "all" in the name.

buildbot/osuosl/master/config/builders.py
```diff
@@ -1848,6 +1824,32 @@
...
+                        "-DLLVM_LIT_ARGS=-v --xunit-xml-output test-results.xml",
+                        '-DLLVM_TARGETS_TO_BUILD=All',
+                        '-DCMAKE_EXPORT_COMPILE_COMMANDS=1',
+                        '-DLLVM_BUILD_LLVM_DYLIB=On',
+                        '-DLLVM_LINK_LLVM_DYLIB=On',
+                        '-DBUILD_SHARED_LIBS=Off',
+                        '-DLLVM_ENABLE_LLD=Off',
+                    ],
+                    env={
+                        'CC': 'ccache gcc', 'CXX': 'ccache g++', 'CCACHE_CPP2': 'yes'
```
gribozavr, line 1850: Please include the PATH as well.

Also, any reason to use gcc instead of clang as the host compiler?

tbaederr: We just discussed this and it is not a requirement for the test, I'll change it to clang.

tstellar:
Could we also have the same configuration added for the release branch?

gribozavr (approved):

tstellar:
@tbaederr We can always add the release branch configuration in a follow up commit, so no need to hold up this patch for it.

gribozavr:
@tbaederr Do you have commit access or do you want me to merge it?

tbaederr:
I can do it, thanks

PR #17, opened 2021-09-14
Fix a capitalisation problem

tbaederr:
The build system doesn't accept "All" here, we must pass "all".

This broke e.g. https://lab.llvm.org/buildbot/#/builders/196/builds/846. The error message is a bit misleading but this works when we pass "all" instead.


tbaederr:
Looks good, @gribozavr?

gribozavr (approved):

PR #18, opened 2021-09-16
Add a llvm-x86_64-debian-dylib mail notifier

tbaederr:
Notify me and @tstellar when the `llvm-x86_64-debian-dylib` builder fails


Meinersbur:
I think the primary place for reviews for llvm-zorg is also reviews.llvm.org.

tbaederr:
Alright, sorry. Opened https://reviews.llvm.org/D109947 instead.

PR #19, opened 2021-10-04
Disable bindings on x86_64-debian-dylib builder

tbaederr:
The tests for the OCaml bindings fail when using the llvm.so dylib.

@gribozavr Does this look right to you?


gribozavr (approved):

gribozavr (changes requested):
Are you editing the right builder? It is not `x86_64-debian-dylib`.

tbaederr:
You are right of course, sorry. Fixed that.

gribozavr (approved):
Do you have permissions to merge?

PR #20, opened 2021-12-17
Add BOLT buildbot

aaupov:
Add BOLT build worker and custom builder to zorg.

Builder is expected to check out binary tests from another repo, currently at https://github.com/rafaelauler/bolt-tests. 
The plan is to transfer the repository over to llvm organization in the future.

With that requirement, BOLT builder has to define an intermediate checkout step between `getLLVMBuildFactoryAndSourcecodeSteps` and `addCmakeSteps`/`addNinjaSteps`. 
Otherwise, it's based on `UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory`.

The configuration is tested with a [temporary build master](http://157.230.108.44:8011/#/builders/3/builds/2811), which points to the [current BOLT repo](https://github.com/facebookincubator/bolt) instead of the monorepo.

This PR should only be merged once BOLT is pushed to the monorepo. 
Current plan is end of this week: https://lists.llvm.org/pipermail/llvm-dev/2021-December/154290.html

### DIFF
--- a/zorg/buildbot/builders/BOLTBuilder.py
+++ b/zorg/buildbot/builders/BOLTBuilder.py
@@ -1,0 +1,69 @@
+from buildbot.plugins import steps
+from zorg.buildbot.commands.CmakeCommand import CmakeCommand
+from zorg.buildbot.builders.UnifiedTreeBuilder import getLLVMBuildFactoryAndSourcecodeSteps, addCmakeSteps, addNinjaSteps
+from zorg.buildbot.process.factory import LLVMBuildFactory
+
+def getBOLTCmakeBuildFactory(
+           clean = False,
+           bolttests = False,
+           extra_configure_args = None,
+           env = None,
+           **kwargs):
+
+    if env is None:
+        env = dict()
+
+    bolttests_dir = "bolt-tests"
+
+    cleanBuildRequested = lambda step: clean or step.build.getProperty("clean", default=step.build.getProperty("clean_obj"))
+
+    checks = ['check-bolt']
+    extra_steps = []
+
+    f = getLLVMBuildFactoryAndSourcecodeSteps(
+            depends_on_projects=['clang', 'lld', 'bolt'],
+            **kwargs) # Pass through all the extra arguments.
+
+    if bolttests:
+        checks += ['check-large-bolt']
+        extra_configure_args += [
+            '-DLLVM_EXTERNAL_PROJECTS=bolttests',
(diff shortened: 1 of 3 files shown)


zorg/buildbot/builders/BOLTBuilder.py
```diff
@@ -0,0 +1,69 @@
...
+
+    bolttests_dir = "bolt-tests"
+
+    cleanBuildRequested = lambda step: clean or step.build.getProperty("clean", default=step.build.getProperty("clean_obj"))
+
+    checks = ['check-bolt']
+    extra_steps = []
+
+    f = getLLVMBuildFactoryAndSourcecodeSteps(
+            depends_on_projects=['clang', 'lld', 'bolt'],
```
joker-eph, line 24: Curious: why does bold depends on clang?

aaupov: BOLT itself doesn't, it's the C/C++ tests that come with it. My understanding is that `depends_on_projects` translate into `-DLLVM_ENABLE_PROJECTS`.

joker-eph: > My understanding is that depends_on_projects translate into -DLLVM_ENABLE_PROJECTS.

No, this is specified separately. This is about what the bot will be triggered on (i.e. changes to clang itself will trigger new builds), to avoid triggering spurious build.

aaupov: Thanks for clarification! Although our C/C++/assembly tests depend on clang and lld, I think it doesn't make sense to trigger BOLT build/test on commits to these projects. I'll drop this `depends_on_projects` line, to reflect that we only depend on llvm which is added by default.

gkistanova:
Hello Amir.

Could you use reviews.llvm.org for your patch review, please. See https://llvm.org/docs/DeveloperPolicy.html#making-and-submitting-a-patch for more details.

We do not use github pull requests at the moment.

As of the patch itself, at a quick glance you may want to take a closer look at the "clean", and "clean_obj" build properties and "clean" arg support.  You might not want to checkout the BOLT test suite from scratch, unless the "clean" build property is set. The "clean_obj" build property and "clean" arg request only build directories / objects to be removed, not a source code.

aaupov:
Hello Galina,
Thank you for the feedback! I'll move the PR over to phabricator.

aaupov:
Created [D116060](https://reviews.llvm.org/D116060) and [D116062](https://reviews.llvm.org/D116062).

PR #21, opened 2022-01-14
Remove fedora builder and worker

kwk:
This patch removes this builder:

 * x86_64-fedora-clang

and this worker:

 * fedora-llvm-x86_64

This builder and worker haven't been maintained by me for a while and
I initially installed them with the intend to support them much longer.
But this was only the beginning of my journey with LLVM CI...

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2089,36 +2089,6 @@ all += [
                     script="ve-linux.py",
                     depends_on_projects=['llvm', 'clang', 'compiler-rt', 'libcxx'])},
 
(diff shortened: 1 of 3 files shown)


gkistanova:
Hi Konrad,

Thanks for the patch!

We are still using Phabricator for code reviews, not github. :)
However, this patch looks good. So, please feel free to just commit directly to llvm-zorg.

kwk:
> Hi Konrad,
> 
> Thanks for the patch!
> 
> We are still using Phabricator for code reviews, not github. :) However, this patch looks good. So, please feel free to just commit directly to llvm-zorg.

Done with 594f09aef2af36bd1609e1c80bf8fc9064869b6f.

PR #25, opened 2022-03-29
Add LIBC_INCLUDE_BENCHMARKS to fullbuild

gchatelet:
This is fixing the build bots now that https://reviews.llvm.org/D122481 has been submitted.

### DIFF
--- a/zorg/buildbot/builders/annotated/libc-linux.py
+++ b/zorg/buildbot/builders/annotated/libc-linux.py
@@ -49,6 +49,7 @@ def main(argv):
 
         if fullbuild and not args.asan:
             cmake_args.append('-DLLVM_LIBC_INCLUDE_SCUDO=ON')
+            cmake_args.append('-DLIBC_INCLUDE_BENCHMARKS=ON')
 
         if fullbuild:
             cmake_args.extend(['-DLLVM_LIBC_FULL_BUILD=ON']),


sivachandra:
Can you please consider this approved and commit this directly? We use Phabricator for zorg also and reviews work exactly the same way if you want a review.

sivachandra:
I have submitted this separately to get the bot to cycle green.

gchatelet:
> Can you please consider this approved and commit this directly? We use Phabricator for zorg also and reviews work exactly the same way if you want a review.

Thx for submitting this while I was away.
I wasn't aware of the review process for zorg. Next time I'll submit this directly. Two more questions:
 - Is it sufficient to just push the commit to fix the issue (aka. are build bot fetching llvm-zorg at HEAD) or is there an additional process to make llvm-zorg available to build bots?
 - It seems that flags for llvm libc should be of the form `LLVM_LIBC_XXX`, do you want me to rename `LIBC_INCLUDE_BENCHMARKS` into `LLVM_LIBC_INCLUDE_BENCHMARKS` for consistency?

sivachandra:
Yes, it is sufficient to push. The bots update the build scripts right at the beginning. For example, look at step 2 here: https://lab.llvm.org/buildbot/#/builders/163/builds/15168.

Normally, the `LLVM_` prefix is used for vars which represent something in the `llvm` directory or something `llvm-project` wide. For vars which affect something only in the `libc` directory, we should not have the `LLVM_` prefix. Take for example the vars in `clang` directory which affect only clang- they do not have a `LLVM_CLANG_` prefix.

PR #28, opened 2022-07-13
[libc++] Stop building libc++ alongside Clang on GreenDragon bots

ldionne:
The libc++ headers used to be shipped alongside the Clang toolchain
on Apple platforms, so it made sense to build libc++ at the same time
as Clang in our CI. However, for a couple of years now, Clang and libc++
are entirely separate projects w.r.t. how they are shipped on Apple
platforms, and it doesn't make sense to build libc++ at the same time
as Clang anymore. Libc++ has its own CI infrastructure.


shahmishal:
LGTM! Thanks

PR #31, opened 2022-08-16
[lldb-cmake-matrix] Use RUNTIME option for libcxx/libcxxabi

felipepiovezan:
As of D112724, building libcxx/libcxxabi with the "PROJECTS" option is
deprecated and causes the following CMake warning:

> Using LLVM_ENABLE_PROJECTS=libcxxabi is deprecated now, please use
-DLLVM_ENABLE_RUNTIMES=libcxxabi or see the instructions at
https://libcxx.llvm.org/BuildingLibcxx.html for building the
runtimes.

This commit fixes the warning, also in the hopes of fixing the failures
in lldb-cmake-matrix that test different clang versions. These older
clangs fail to find the correct libcxx, instead using the host machine's
libcxx.


felipepiovezan:
moved to phab: https://reviews.llvm.org/D131970

PR #37, opened 2023-06-03
Build more projects with llvm-x86_64-debian-dylib

tstellar:
We want to expand the dylib testing to include libclang-cpp.so and also to build projects like lld, lldb, and clang-tools-extra that consume the dylibs.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2324,8 +2324,8 @@ all += [
     'builddir': 'llvm-x86_64-debian-dylib',
     'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
                     clean=True,
-                    depends_on_projects=['llvm'],
-                    checks=['check-all'],
+                    depends_on_projects=['llvm', 'clang', 'lldb', 'lld', 'clang-tools-extra'],
+                    checks=['check-clang', 'check-llvm', 'check-lld', 'check-clang-extra'],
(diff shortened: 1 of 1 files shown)


tstellar:
@gkistanova Is there some way to test this change out in staging before it is pushed to the production server?

buildbot/osuosl/master/config/builders.py
```diff
@@ -2324,8 +2324,8 @@
     'builddir': 'llvm-x86_64-debian-dylib',
```
joker-eph, line 2324: Shall the builder be renamed to reflect its new scope?

tstellar: I thought about that, but wasn't sure exactly what to call it? llvm-project-x86_64-dylib?

gkistanova:
@tstellar You can either commit, wait for a couple of hours, and connect the worker to staging. This way you will be able to see how it works before the changes will go to the production. Or if anything in particular worries you, talk to me and I'll stage the patch for you without committing. usually committed changes go to production on Fridays or over a weekend.

The patch itself looks good. But you need to decide between #37 and #39. If I get it right, these 2 patches are mutually exclusive?

tstellar:
> The patch itself looks good. But you need to decide between #37 and #39. If I get it right, these 2 patches are mutually exclusive?

I don't think they are necessarily mutually exclusive.  @serge-sans-paille is saying we should use both configs.

gkistanova (approved):
LGTM.

PR #44, opened 2023-09-07
Add builders for premerge-like configurations

metaflow:
As was suggested https://github.com/llvm/llvm-project/issues/56674#issuecomment-1690845357 that adds a configuration that should match build run by GitHub pull requests.

We are going to set up initially one worker per windows / linux architecture. It's unlikely that windows one will be able to catch up with every commit but we will monitor it and decide if e.g. adding one more worker might actually be fast enough to build every change.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2792,4 +2792,45 @@ all += [
                     env={
                         'CCACHE_DIR' : WithProperties("%(builddir)s/ccache-db"),
                     })},
+
+    # Builders similar to used in Buildkite premerge pipeline.
+    # Please keep in sync with llvm-project/.ci configurations.
+
+    # See https://github.com/llvm/llvm-project/blob/main/.ci/monolithic-windows.sh.
+    {'name' : "premerge-monolithic-windows",
+    'tags'  : ["premerge"],
+    'collapseRequests': True,
+    'workernames' : ["premerge-windows-1"],
+    'builddir': "premerge-monolithic-windows",
+    'factory' : UnifiedTreeBuilder.getCmakeWithNinjaWithMSVCBuildFactory(
+                    vs="autodetect",
+                    depends_on_projects=["clang-tools-extra", "clang", "flang", "libclc", "lld", "llvm", "mlir", "polly", "pstl"],
+                    checks=["check-all"],
+                    extra_configure_args=[
+                        "-DCMAKE_BUILD_TYPE=Release",
+                        "-DLLVM_ENABLE_ASSERTIONS=ON",
+                        "-DLLVM_BUILD_EXAMPLES=ON",
+                        "-DCOMPILER_RT_BUILD_LIBFUZZER=OFF",
+                        "-DLLVM_LIT_ARGS=-v",
+                        "-DCOMPILER_RT_BUILD_ORC=OFF",
+                        "-DCMAKE_C_COMPILER_LAUNCHER=sccache",
+                        "-DCMAKE_CXX_COMPILER_LAUNCHER=sccache"])},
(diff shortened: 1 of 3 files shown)


aeubanks:
Having reviewed https://github.com/llvm/llvm-project/pull/65574/files, I'm confused as to the "scheduled" vs "premerge" here. I thought we'd be using the "scheduled" for this.

joker-eph:
Thanks for adding this!

Buildbot is the only system that notifies post-commit failure right now, the buildkite "scheduled" flow is useful for people monitoring explicitly buildkite, not for the community at large.

(LGTM, but @gkistanova is the main reviewer for this!)

metaflow:
> Having reviewed https://github.com/llvm/llvm-project/pull/65574/files, I'm confused as to the "scheduled" vs "premerge" here. I thought we'd be using the "scheduled" for this.

distinction between "scheduled" and "premerge" makes sense in .ci but here it's mostly about highlighting the fact that the configuration used is matching whatever we have for Pull Request testing.

buildbot/osuosl/master/config/builders.py
```diff
@@ -2792,4 +2792,45 @@
...
                         'CCACHE_DIR' : WithProperties("%(builddir)s/ccache-db"),
                     })},
+
+    # Builders similar to used in Buildkite premerge pipeline.
+    # Please keep in sync with llvm-project/.ci configurations.
+
+    # See https://github.com/llvm/llvm-project/blob/main/.ci/monolithic-windows.sh.
+    {'name' : "premerge-monolithic-windows",
+    'tags'  : ["premerge"],
+    'collapseRequests': True,
```
gkistanova, line 2802: If you want to collapse build requests, just skip the `collapseRequests` arg and let the default machinery do the job. Setting it to True is a bad idea, as there is a number of build requests which should not be collapsed, like manual build requests, for example, or requested clean builds, and such.

buildbot/osuosl/master/config/status.py
```diff
@@ -380,4 +380,11 @@ def getReporters():
             subject = "Build %(builder)s Failure",
             mode = "failing",
             builders = ["clang-cmake-x86_64-avx512-linux"]),
+        reporters.MailNotifier(
+            fromaddr = "llvm.buildmaster@lab.llvm.org",
+            sendToInterestedUsers = False,
+            messageFormatter = LLVMInformativeMailNotifier, # TODO: remove after moving to prod.
```
gkistanova, line 386: Could you elaborate what this TODO comment means, please?

gkistanova (changes requested):
Thanks for the patch, Mikhail!

It looks good in general, with a few things to address.

buildbot/osuosl/master/config/builders.py
metaflow: Done.

buildbot/osuosl/master/config/status.py
metaflow: in instructions it is said:

> It is possible to allow email addresses to unconditionally receive notifications on build failure; for this you’ll need to add an InformativeMailNotifier to buildbot/osuosl/master/config/status.py. This is particularly useful for the staging buildmaster which is silent otherwise.

So I thought it's a good idea to have it but only for the "staging" phase. So I am planning to replace notification settings with a "common" one afterwards. Can drop this todo altogether if that helps.

metaflow: removed and added a note to myself

metaflow:
@gkistanova I am not sure if you received an update here - PR interface still shows that "1 change requested" even though I seemed to address both comments :/

metaflow:
@gkistanova I am planning to merge this tommorrow if there is no other concerns.

PR #46, opened 2023-09-08
[jenkins] Coverage bot: Remove cross-project-tests

gulfemsavrun:
Directory coverage is recently enabled in https://reviews.llvm.org/D158703 The following file shows up at the top level because cross-project-tests is also built in the coverage build:
`cross-project-tests/debuginfo-tests/llvm-prettyprinters/gdb/llvm-support.cpp`
This patch removes cross-project-tests from the project list.

Differential Revision: https://reviews.llvm.org/D159469

### DIFF
--- a/zorg/jenkins/jobs/jobs/llvm-coverage
+++ b/zorg/jenkins/jobs/jobs/llvm-coverage
@@ -76,7 +76,7 @@ pipeline {
 
                     python llvm-zorg/zorg/jenkins/monorepo_build.py lldb-cmake build \
                       --assertions \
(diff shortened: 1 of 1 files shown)


yhgu2000:
I don't know if disabling the cross-project-test affects the coverage metrics. Will it be better if we use the [`--ignore-filename-regex`](https://llvm.org/docs/CommandGuide/llvm-cov.html#cmdoption-llvm-cov-show-ignore-filename-regex) optior to just hide it in the report? In that case, we need to modify the Python script instead.

PR #47, opened 2023-09-14
[libc++] Don't build libc++ benchmarks in the `publish-runtimes-sphinx-docs` job

var-const:
The immediate motivation is to fix the job (which currently fails because the environment doesn't support C++23 which the benchmarks require to compile), but building benchmarks should be unnecessary to publish the docs anyway.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2149,7 +2149,11 @@ all += [
     'tags'  : ["doc"],
     'workernames' : ["as-worker-4"],
     'builddir': "publish-runtimes-sphinx-docs",
-    'factory' : SphinxDocsBuilder.getLLVMRuntimesDocsBuildFactory(clean=True)},
+    'factory' : SphinxDocsBuilder.getLLVMRuntimesDocsBuildFactory(
+                    clean=True,
+                    extra_configure_args=[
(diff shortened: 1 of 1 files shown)


gkistanova (approved):
Thanks, Konstantin!
LGTM.

ldionne:
@gkistanova Do you know why this change apparently never made it to the builder? From https://lab.llvm.org/buildbot/#/builders/242/builds/1130/steps/4/logs/stdio:

```
cmake ../llvm-project/runtimes -DLLVM_ENABLE_SPHINX=ON -DSPHINX_OUTPUT_HTML=ON -DSPHINX_OUTPUT_MAN=OFF -DSPHINX_WARNINGS_AS_ERRORS=OFF -DLLVM_ENABLE_ASSERTIONS=OFF -DCMAKE_BUILD_TYPE=Release '-DLLVM_ENABLE_RUNTIMES=libc;libcxxabi;libunwind;libcxx' '-DLLVM_LIT_ARGS=-v -vv' -GNinja
```

This is a job that ran today, and it doesn't contain the flag added by this patch.

var-const:
@gkistanova Another question -- would it be possible to update the compiler version in the builder? The root cause here is that the builder uses a version of GCC that we no longer support (I believe it's `9.4`) and that doesn't support C++23. FWIW, the minimum GCC version supported by libc++ is `12`.

PR #48, opened 2023-09-27
Add two new Sony builders and workers

dyung:
Add two new builders clang-x86_64-linux-abi-test and llvm-new-debug-iterators,
two new workers sie-linux-worker2 and sie-linux-worker3 as well as setup notifications for their failures.

I'm not sure what configuration we will use for the machines yet and wanted to get them into the staging server so that I can see what the timings are like and what state things are in at the moment before I commit to a configuration.


dyung:
ping

gkistanova (approved):
LGTM.

PR #56, opened 2023-10-12
[clang-ppc64-aix] Remove absolute paths

jakeegan:
This makes it more generic so we can update clang or python without updating the paths in this repo.


daltenty (approved):
LGTM, thanks!

gkistanova (approved):
LGTM.

PR #57, opened 2023-10-13
Make builder llvm-new-debug-iterators do clean builds.

dyung:
Set option to have the build bot do clean builds.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -978,6 +978,7 @@ all = [
     'builddir': "debug-iterators",
     'factory': UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
                     depends_on_projects=['llvm','clang','clang-tools-extra','compiler-rt','lld','cross-project-tests'],
+                    clean=True,
                     extra_configure_args=[
                         "-DCMAKE_C_COMPILER=gcc",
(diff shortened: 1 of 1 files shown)


gkistanova (approved):
LGTM.

joker-eph:
There is no motivation exposed here: can you clarify on why this change?

dyung:
> There is no motivation exposed here: can you clarify on why this change?

The bot was previously doing "dirty" builds while it was in testing and waiting for an upstream commit to go in. Now that the commit is in and it added a new CMake argument, I was afraid that doing dirty builds may not get the CMake argument to stick so we wouldn't be testing what we were hoping to test with this bot.

In addition, ccache is enabled on this bot and should be roughly equivalent time wise to doing dirty builds, so it didn't feel like a bad thing to enable.

Hope that helps to explain the motivation behind this change.

joker-eph:
> I was afraid that doing dirty builds may not get the CMake argument to stick so we wouldn't be testing what we were hoping to test with this bot.

So there is no evidence of anything wrong right now? It's just not clear why this bot in particular who be more sensitive to incremental build issue than any other?

> If you'd like to keep your personal email address private, you can use a noreply email address from GitHub as your commit email address. To use your noreply email address for commits you push from the command line, use that email address when you set your commit email address in Git. To use your noreply address for web-based Git operations, set your commit email address on GitHub and choose to Keep my email address private.

PR #60, opened 2023-10-20
[BOLT] Build lld to avoid test breakage

aaupov:
Use trunk lld in BOLT tests instead of BOLT_LLD_EXE to reduce
maintenance burden. lld is cheap to build, and it helps to track lld
commits too.


gkistanova (approved):
LGTM.

PR #61, opened 2023-10-23
Add notification for Jeremy Morse for llvm-new-debug-iterators builder.

dyung:
Jeremy Morse wants to be notified for failures of the llvm-new-debug-iterators buildbot as it is a configuration he is interested in.


gkistanova (approved):
LGTM.

PR #66, opened 2023-11-03
[Android] Remove unneeded -stdlib and -I flags for C++

rprichard:
A recent Clang driver will assume -stdlib=libc++ automatically for Android and will also automatically add $SYSROOT/usr/include/c++/v1 to the include search path.

This explicit -I path breaks libFuzzer when it builds and links its own copy of libc++ because there are two different libc++ copies on the include path.


petrhosek (approved):

vitalybuka (approved):

vitalybuka:
It does not help.
Note, you can run that script on local workstation, from empty dir.
It will run tests on attached phones, but it should build without any attached devices just fine

vitalybuka:
https://lab.llvm.org/buildbot/#/builders/77/builds/31924

rprichard:
> https://lab.llvm.org/buildbot/#/builders/77/builds/31924

I think this is an unrelated build failure:

```
/var/lib/buildbot/sanitizer-buildbot6/sanitizer-x86_64-linux-android/build/llvm-project/llvm/lib/Transforms/IPO/MergeFunctionsIgnoringConst.cpp:1271:16: error: 'supportsTypedPointers' is deprecated: Always returns false [-Werror,-Wdeprecated-declarations]
 1271 |   if (!context.supportsTypedPointers())
      |                ^
```

This seems to be the fix, https://github.com/llvm/llvm-project/pull/71219.

vitalybuka:
Sorry, you are correct.
Thanks for the fix!

PR #69, opened 2023-11-07
[ClangBuilder] Fix compiler-rt LLVM_CONFIG_PATH is deprecated warning

c-rhodes:
This warning is currently being emitted by builders that build compiler-rt, e.g.

  CMake Warning at
  /home/tcwg-buildbot/worker/clang-aarch64-full-2stage/llvm/compiler-rt/cmake/Modules/CompilerRTUtils.cmake:281
  (message):
    LLVM_CONFIG_PATH is deprecated, please use LLVM_CMAKE_DIR instead

this patch sets LLVM_CMAKE_DIR to remove it.

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -341,14 +341,17 @@ def _getClangCMakeBuildFactory(
             ])
 
     rel_src_dir = LLVMBuildFactory.pathRelativeTo(f.llvm_srcdir, stage1_build)
+    cmake_cmd1 = [cmake, "-G", "Ninja", rel_src_dir,
+                  "-DCMAKE_BUILD_TYPE="+stage1_config,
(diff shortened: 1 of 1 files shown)


gkistanova:
With all due respect, this looks like a wrong place to fix the compiler-rt cmake issue. LLVM_CMAKE_DIR is not supposed to be set manually. That's a part of the LLVM cmake machinery. Unless I'm missing something, of course.

c-rhodes:
> With all due respect, this looks like a wrong place to fix the compiler-rt cmake issue. LLVM_CMAKE_DIR is not supposed to be set manually. That's a part of the LLVM cmake machinery. Unless I'm missing something, of course.

Thanks for your comment, apologies I missed it last week. Closing this as it's not the correct fix, not sure why LLVM_CMAKE_DIR isn't set in buildbot, I may take another look when I get chance.

PR #72, opened 2023-11-15
Use annotation protocol for the Fuchsia builder

petrhosek:
This streamlines development as we can deploy changes without having to restart the Buildbot master.


rnk (commented):
+1 for more annotation builders

gkistanova (approved):
LGTM.

PR #75, opened 2023-11-28
[TestSuiteBuilder] Updating the test-suite clobber step

kamaub:
This change modifies the step that clobbers the build directory of the llvm-test-suite to use the standardized
`f.addStep(steps.RemoveDirectory())` that the base class, UnifiedTreeBuilder, uses instead of the more explicit `f.addStep(ShellCommand(command=['rm', '-rf']))`

### DIFF
--- a/zorg/buildbot/builders/TestSuiteBuilder.py
+++ b/zorg/buildbot/builders/TestSuiteBuilder.py
@@ -15,6 +15,7 @@ def addTestSuiteStep(
             f,
             compiler_dir = None,
             env = None,
+            cleanBuildRequested=True,
             lit_args = [],
             **kwargs):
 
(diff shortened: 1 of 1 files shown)


zorg/buildbot/builders/TestSuiteBuilder.py
```diff
@@ -28,13 +29,16 @@ def addTestSuiteStep(
...
-                           env=env))
+    # The default value of cleanBuildRequested is TRUE as we should always
+    # clobber the build directory to test each freshly built compiler.
+    f.addStep(steps.RemoveDirectory(
+                name='Clean Test Suite Build dir' % test_suite_workdir,
+                dir=test_suite_workdir,
+                haltOnFailure=False,
+                flunkOnFailure=False,
+                doStepIf=cleanBuildRequested,
+                ))
```
lei137, line 40: nit: Should we keep to the same indentation here as other files? eg .
```suggestion
    f.addStep(steps.RemoveDirectory(name='Clean Test Suite Build dir' % test_suite_workdir,
              dir=test_suite_workdir,
              haltOnFailure=False,
              flunkOnFailure=False,
              doStepIf=cleanBuildRequested,
              ))
```

lei137 (approved):
LGTM
Minor indentation nit.

zorg/buildbot/builders/TestSuiteBuilder.py
```diff
@@ -28,13 +29,16 @@ def addTestSuiteStep(
...
-    f.addStep(ShellCommand(name='Clean Test Suite Build dir',
-                           command=['rm', '-rf', test_suite_workdir],
-                           haltOnFailure=True,
-                           description='Removing the Test Suite build directory',
-                           workdir=test_suite_base_dir,
-                           env=env))
+    # The default value of cleanBuildRequested is TRUE as we should always
+    # clobber the build directory to test each freshly built compiler.
+    f.addStep(steps.RemoveDirectory(
+                name='Clean Test Suite Build dir' % test_suite_workdir,
```
gkistanova, line 35: Could you use Python3 `.format()` here instead, please?

gkistanova (commented):
The patch looks fine with a small thing to fix.

zorg/buildbot/builders/TestSuiteBuilder.py
amy-kwan: I think the indenting would probably start under `name=`, like the other steps.

amy-kwan (approved):
No further comments from me once previous comments have been addressed.

PR #76, opened 2023-11-30
Change fuchsia annotated builder to use getcwd() instead of "."

trevtt:
This change fixes the fuchsia annotated builder to work in a Docker container.


zeroomega:
LGTM

PR #80, opened 2023-12-05
Upgrade to Clang 17.0.6 in the C++20 buildbot

ilya-biryukov:
See https://github.com/llvm/llvm-project/pull/72348 for some context on why we need this upgrade.

### DIFF
--- a/buildbot/google/docker/buildbot-cpp20/Dockerfile
+++ b/buildbot/google/docker/buildbot-cpp20/Dockerfile
@@ -18,15 +18,16 @@ RUN apt-get update && \
         cmake ninja-build  \
         ccache \
         python3 python3-pip python3-psutil && \
-    # Install clang-15 to make the minimal C++20 support in the toolchain.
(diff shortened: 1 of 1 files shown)


ilya-biryukov:
I have checked that this works, but I believe we should wait until the [RFC](https://discourse.llvm.org/t/rfc-clang-17-0-6-would-be-minimum-version-to-build-llvm-in-c-20/75345/12) gets discussed properly before landing this.

Putting Aaron as the reviewer for this change to stamp when the discussion has reached some conclusion.

AaronBallman (approved):
LGTM assuming there are no surprises on the RFC in the next few weeks.

usx95 (approved):
LGTM. [As Aaron mentioned on the RFC](https://discourse.llvm.org/t/rfc-clang-17-0-6-would-be-minimum-version-to-build-llvm-in-c-20/75345/14), we will wait for two more weeks for comments on the RFC.
If no concerns are raised, we can merge this on 21-Dec-2023.

PR #82, opened 2023-12-12
Dockerfile for Fuchsia builders

petrhosek:
This Dockerfile is used to build a Docker image that's used on all Fuchsia builders. Our infrastructure will automatically pick up any changes to this Dockerfile, rebuild and redeploy the image.


zeroomega (approved):

gkistanova (approved):
LGTM.

PR #86, opened 2023-12-28
[libc][AMDGPU] Enable parallel tests for libc on AMDGPU

jplehr:
It appears that https://github.com/llvm/llvm-project/pull/70695 fixed an issue that prevented libc unit tests to be run in parallel.


ronlieb (approved):

jhuber6:
Should've mentioned, this default to 1 now since it was always subtly broken with full parallelism. (I got like one in 1000 runs to deadlock when run on CUDA on like 48 threads, so figured it was better safe that sorry).

https://github.com/llvm/llvm-project/commit/a28e4eac26ccd2e17f041b8d96da93df4a2414c0 <- relevant patch.

jplehr:
Oh, I see.
I had that variable set to something like 64 in one terminal and not set in another and simply confused the two terminals. I'll do some more testing locally to see if I can trigger the deadlock you mentioned and if not, I'll put up another PR.

PR #89, opened 2024-01-04
Fix fuchsia annotated build path

trevtt:
The fuchsia build dir is set to the current working directory. 
This change moves build dir  to be a sibling of the llvm-project source directory so it can be safely cleaned for each build.

### DIFF
--- a/zorg/buildbot/builders/annotated/fuchsia-linux.py
+++ b/zorg/buildbot/builders/annotated/fuchsia-linux.py
@@ -4,6 +4,7 @@ import argparse
 import os
 import subprocess
 import sys
+import tempdir
 import traceback
 import util
 
@@ -41,12 +42,8 @@ def main(argv):
     buildbot_revision = os.environ.get('BUILDBOT_REVISION', 'origin/main')
 
     cwd = os.getcwd()
-    build_dir = os.path.join(cwd, '..', 'llvm-build')
+    build_dir = tempfile.mkdtemp('', 'llvm-build-', cwd)
     source_dir = os.path.join(cwd, '..', 'llvm-project')
-
-    if buildbot_clobber:
-        with step('clean', halt_on_fail=True):
-            util.clean_dir(build_dir)
 
     with step('configure', halt_on_fail=True):
         cmake_args = [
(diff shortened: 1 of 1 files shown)


zorg/buildbot/builders/annotated/fuchsia-linux.py
```diff
@@ -41,7 +41,7 @@ def main(argv):
     buildbot_revision = os.environ.get('BUILDBOT_REVISION', 'origin/main')
 
     cwd = os.getcwd()
-    build_dir = cwd
+    build_dir = os.path.join(cwd, '..', 'llvm-build')
```
petrhosek, line 44: I'd suggest keeping the build directory under the CWD:
```suggestion
    build_dir = os.path.join(cwd, 'llvm-build')
```

zorg/buildbot/builders/annotated/fuchsia-linux.py
```diff
@@ -41,13 +42,9 @@ def main(argv):
     buildbot_revision = os.environ.get('BUILDBOT_REVISION', 'origin/main')
 
     cwd = os.getcwd()
-    build_dir = os.path.join(cwd, '..', 'llvm-build')
+    build_dir = tempfile.mkdtemp('', 'llvm-build-', cwd)
     source_dir = os.path.join(cwd, '..', 'llvm-project')
 
-    if buildbot_clobber:
-        with step('clean', halt_on_fail=True):
-            util.clean_dir(build_dir)
```
petrhosek, line 49: I think it might be still worth supporting this, but rather than cleaning the `build_dir`, it should empty out `cwd` by calling `util.clean_dir(cwd)`, but this needs to happen before calling `tempfile.mkdtemp`.

zorg/buildbot/builders/annotated/fuchsia-linux.py
```diff
@@ -41,13 +42,9 @@ def main(argv):
     buildbot_revision = os.environ.get('BUILDBOT_REVISION', 'origin/main')
 
     cwd = os.getcwd()
-    build_dir = os.path.join(cwd, '..', 'llvm-build')
+    build_dir = tempfile.mkdtemp('', 'llvm-build-', cwd)
```
petrhosek, line 45: I'd use keyword arguments to make it clear what each argument is.
```suggestion
    build_dir = tempfile.mkdtemp(prefix='llvm-build-', dir=cwd)
```

zorg/buildbot/builders/annotated/fuchsia-linux.py
petrhosek: We should also update the ccache configuration to place the cache in `cwd` but that can be done in a follow up change.

PR #92, opened 2024-01-05
[LibC][AMDGPU] Enable 8-way parallel AMDGPU tests

jplehr:
This enables an actual 8-way parallel testing of libc for GPU on AMDGPU. We limit the number of concurrent executions since we have some anecdotal evidence that highly-parallel execution can still provoke some issues. I think we should test some parallel execution though and we did not see hangs on other systems with low numbers of parallel jobs.


jhuber6 (approved):
Sure, why not. High parallelism can also exhaust resources, since if you have like 64 kernels launching all at once it's going to be using a lot of memory / queues.

jplehr:
Let's see if it blows up or not and how much faster turnaround time we get with this change.

PR #94, opened 2024-01-08
[OpenMP] Adds another AMDGPU OpenMP bot

jplehr:
This adds a sanity-check bot to the AMDGPU OpenMP Offloading bot infrastructure. It is a copy of the existing production bot, but with an updated OS and ROCm base-installation and running on a more potent machine.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1793,6 +1793,32 @@ all += [
                         ],
                     )},
 
+    {'name' : "openmp-offload-amdgpu-runtime-2",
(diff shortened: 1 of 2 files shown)


jhuber6:
Unrelated, but I've wondered if we should have a public bot that always builds with the `dlopen` HSA enabled.

jplehr:
> Unrelated, but I've wondered if we should have a public bot that always builds with the `dlopen` HSA enabled.

I don't have a strong opinion on this. We mainly need to see what sort of resources we can allocate for what testing, can discuss this internally.

gkistanova (approved):
LGTM.

PR #95, opened 2024-01-08
[Flang] Adds mlir explicitly and runs its tests

jplehr:
Flang depends on MLIR, so we enable the component explicitly and add the check-mlir lit tests to be run.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1831,7 +1831,7 @@ all += [
     'factory' : OpenMPBuilder.getOpenMPCMakeBuildFactory(
                         clean=True,
                         enable_runtimes=['openmp'],
-                        depends_on_projects=['llvm','clang','lld','openmp','flang'],
+                        depends_on_projects=['llvm','clang','lld','openmp','mlir','flang'],
                         extraCmakeArgs=[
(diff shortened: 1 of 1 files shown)


jhuber6 (approved):

jplehr:
I'm gonna run this through some more offline testing first.

joker-eph:
If you’re validating flang, I don’t understand why you need to run MLIR tests? (Or why is it useful)

jhuber6:
> If you’re validating flang, I don’t understand why you need to run MLIR tests? (Or why is it useful)

My understanding is that `Flang` uses MLIR as its intermediate language, so testing it is valuable. Perhaps you are suggesting that such testing is redundant because `flang` uses it implicitly?

jplehr:
> If you’re validating flang, I don’t understand why you need to run MLIR tests? (Or why is it useful)

Yes, my understanding was that the Flang team also adds MLIR tests in their development, which motivated this change. If this is not helpful, since it's already covered / redundant, I can close the PR.
We actually had a bit more discussion internally and thus want to evaluate the impact of this change on turnaround time locally first.

gkistanova:
I converted this PR to draft. Please feel free to either mark it ready for review or close depending on your impact evaluation.

PR #98, opened 2024-01-12
Remove "jobs" property from Linaro workers

DavidSpickett:
This property is used for `ninja -j<jobs>` and given that we manage the CPU usage of our machines using containers or dedicated machines per worker, it's not needed.

Many of them weren't getting this many cores anyway so it should not change much.

The lldb arm bot does have a -j set for the tests, but this is done in the builder configuration instead. The test suite is fine on many cores, but not when there are many other containers trying to use them as well.


omjavaid (approved):
LGTM. Thanks!

PR #100, opened 2024-01-13
Remove BOLT_CLANG_EXE from premerge builder

aaupov:
The option doesn't make sense as clang is being built and BOLT tests can
use trunk clang instead.


gkistanova (approved):
LGTM.

PR #102, opened 2024-01-18
[LLDB] Remove builder class jobs option, only use worker's jobs property if present

DavidSpickett:
Previously the builder class argument "jobs" defaulted to a string to use with `WithProperties`. This assumed that every worker has a "jobs" property, which is no longer true after changes to Linaro's workers.

The behaviour of `WithProperties` is detailed here: https://github.com/buildbot/buildbot/blob/6e7218792bc324584af980bd1abb4d8b6a5055b8/master/docs/manual/configuration/properties.rst#withproperties

As no builder was using the `jobs=` argument I have removed it and adapted code from `NinjaCommand` to add `-j<jobs>` only if the property exists.

Meaning that Linaro's workers that do not have a "jobs" property will not be limited. `lldb-x86_64-debian` will have a `-j` because its worker specifies `'jobs': 72`.

According to Buildbot docs, `WithProperties` should be replaced with `Iterpolate`, but I think doing so here would only complicate this change.

### DIFF
--- a/zorg/buildbot/builders/LLDBBuilder.py
+++ b/zorg/buildbot/builders/LLDBBuilder.py
@@ -11,7 +11,9 @@ from zorg.buildbot.process.factory import LLVMBuildFactory
 # CMake builds
 def getLLDBCMakeBuildFactory(
             clean=False,
-            jobs="%(jobs)s",
+            # None means use the "jobs" property of the worker, if it has one.
+            # If it is set, it should be set to the number of jobs to use.
+            jobs=None,
 
             # Source directory containing a built python
             python_source_dir=None,
@@ -50,9 +52,18 @@ def getLLDBCMakeBuildFactory(
     lit_args = '-v'
 
     if jobs:
-        build_cmd.append(WithProperties("-j%s" % jobs))
-        install_cmd.append(WithProperties("-j%s" % jobs))
-        test_cmd.append(WithProperties("-j%s" % jobs))
+        jobs_option = ["-j{}".format(jobs)]
+    else:
+        # Use the worker property if option was not explicitly specified.
+        jobs_option = [
+            # If jobs exists, add -j. If not, add nothing.
+            WithProperties("%(jobs:+-j)s"),
+            # If jobs exists, add its value. If not, add nothing.
+            WithProperties("%(jobs:-)s")]
+
+    build_cmd.extend(jobs_option)
+    install_cmd.extend(jobs_option)
+    test_cmd.extend(jobs_option)
 
     ############# CLEANING
     cleanBuildRequested = lambda step: clean or step.build.getProperty("clean", default=step.build.getProperty("clean_obj"))
--- a/zorg/buildbot/builders/LLDBBuilder.py
+++ b/zorg/buildbot/builders/LLDBBuilder.py
@@ -11,10 +11,6 @@ from zorg.buildbot.process.factory import LLVMBuildFactory
 # CMake builds
 def getLLDBCMakeBuildFactory(
             clean=False,
-            # None means use the "jobs" property of the worker, if it has one.
-            # If this argument is not None, it should be set to the number of
-            # jobs to use.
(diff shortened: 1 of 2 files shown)


DavidSpickett:
Replaces the workaround done in https://github.com/llvm/llvm-zorg/commit/0e478ea0c92c809f0ac9dfba2a6fdb168d0ea0cc.

zorg/buildbot/builders/LLDBBuilder.py
```diff
@@ -50,9 +53,18 @@ def getLLDBCMakeBuildFactory(
     lit_args = '-v'
 
     if jobs:
-        build_cmd.append(WithProperties("-j%s" % jobs))
-        install_cmd.append(WithProperties("-j%s" % jobs))
-        test_cmd.append(WithProperties("-j%s" % jobs))
+        jobs_option = ["-j{}".format(jobs)]
+    else:
+        # Use the worker property if option was not explicitly specified.
```
luporl, line 58: Just some questions to help me understand this part.

Is each builder free to use the worker property `jobs` in the way that best fits it?
Is it ok for this function's jobs to be greater than worker's `jobs`?

zorg/buildbot/builders/LLDBBuilder.py
```diff
@@ -50,9 +53,18 @@ def getLLDBCMakeBuildFactory(
...
+        jobs_option = [
+            # If jobs exists, add -j. If not, add nothing.
+            WithProperties("%(jobs:+-j)s"),
```
luporl, line 61: Does `WithProperties` always get the worker properties? Or is there only one set of properties and `jobs` happen to be set by workers?

luporl (commented):
LGTM.

I've just added some questions to better understand the logic here.

zorg/buildbot/builders/LLDBBuilder.py
DavidSpickett: Good point. If you set jobs > number of cores, the best case is ninja uses the number of cores anyway, worst case you're creating some extra processes that have to wait around. Certainly there's no good reason to do this.

Given that, I think it would be better to remove the `jobs=` parameter and only use the worker property (if present).

zorg/buildbot/builders/LLDBBuilder.py
DavidSpickett: It's the latter. https://github.com/buildbot/buildbot/blob/6e7218792bc324584af980bd1abb4d8b6a5055b8/master/docs/manual/configuration/properties.rst#properties shows all the things that can set properties, and the last one in wins.

```
If the same property is supplied in multiple places, the final appearance takes precedence. For example, a property set in a builder configuration will override the one supplied by the scheduler.
```

So in this case, a build step can override the builder, which can override the worker.

zorg/buildbot/builders/LLDBBuilder.py
DavidSpickett: I've done this and updated the title/commit message to explain that. Much simpler this way.

zorg/buildbot/builders/LLDBBuilder.py
luporl: Thanks for the explanation. This part is now clear to me.

gkistanova (approved):
The patch looks fine.

Instead of dialing with `jobs` and build `ninja` command by yourself, you could use `[NinjaCommand]`(https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/commands/NinjaCommand.py), which would take care of both, `jobs` and `loadaverage` properties.

DavidSpickett:
> you could use [NinjaCommand]

Will look at this for a follow up change.

DavidSpickett:
...and I broke the lldb bots again, should have taken your advice as per usual :)

https://github.com/llvm/llvm-zorg/pull/122 to fix them and move to NinjaCommand in the process.

PR #103, opened 2024-01-19
Switch sparc64-linux to single-stage builds

glaubitz:
As more than 400 tests are currently failing for the Sparc target, it makes little sense to test two-stage builds until these issues have all been sorted out. Thus, switch sparc64 builds on Linux to single stage.


gkistanova (approved):
LGTM.

PR #109, opened 2024-01-26
[LLVMPoller] Fix an exception in case of changing files in the root

slydiman:
`project=None` will cause the following exception
```
File "/zorg/buildbot/changes/llvmgitpoller.py", line 188, in _process_changes
    project=",".join(projects) if projects else "llvm-project",
builtins.TypeError: sequence item 0: expected str instance, NoneType found
```
Use a dummy str instead.

### DIFF
--- a/zorg/buildbot/changes/llvmgitpoller.py
+++ b/zorg/buildbot/changes/llvmgitpoller.py
@@ -70,7 +70,7 @@ class LLVMPoller(changes.GitPoller):
                 continue
 
             pieces = path.split('/')
-            project = pieces[0] if len(pieces) > 1 else None
+            project = pieces[0] if len(pieces) > 1 else 'root'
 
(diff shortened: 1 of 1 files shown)


gkistanova:
Hm. This is strange. If `project` sets to `None` in line 73, it should throw in line 80, unless I'm missing anything. How did you get it to the line 188?

Anyway, your proposed fix should work. Could you make it `'llvm-project'` instead of `'root'`, please?

slydiman:
> it should throw in line 80, unless I'm missing anything

```
d = {}
k = None
d[k] = 'test'
print(d)
```
Output: `{None: 'test'}`

> Could you make it 'llvm-project' instead of 'root', please?

Done.

gkistanova (approved):
LGTM.

PR #110, opened 2024-01-26
[LLVMPoller] Fix invalid repository while creating a new change

slydiman:
LLVMPoller can be configured to use other `repourl` for test/debug purpose. But LLVMPoller always provides `https://github.com/llvm/llvm-project` as the repository while creating a new change. Then `GitHubStatusPush._extract_github_info()` extracts invalid `repo_owner` and `repo_name`. This patch fixes the issue described above.

### DIFF
--- a/zorg/buildbot/changes/llvmgitpoller.py
+++ b/zorg/buildbot/changes/llvmgitpoller.py
@@ -187,6 +187,6 @@ class LLVMPoller(changes.GitPoller):
                        category=self.category, ## TODO: Figure out if we could support tags here
                        project=",".join(projects) if projects else "llvm-project",
                        # Always promote an external github url of the LLVM project with the changes.
-                       repository=self.repourl,
+                       repository=self.repourl if self.repourl.startswith('https://github.com') else self._repourl,
                        src='git', # Must be one of the buildbot.process.users.srcs
                        properties=properties)
--- a/zorg/buildbot/changes/llvmgitpoller.py
+++ b/zorg/buildbot/changes/llvmgitpoller.py
@@ -187,6 +187,6 @@ class LLVMPoller(changes.GitPoller):
                        category=self.category, ## TODO: Figure out if we could support tags here
                        project=",".join(projects) if projects else "llvm-project",
                        # Always promote an external github url of the LLVM project with the changes.
-                       repository=self._repourl,
+                       repository=self.repourl,
                        src='git', # Must be one of the buildbot.process.users.srcs
                        properties=properties)


gkistanova (approved):
Thanks, Dmitry!
LGTM.

gkistanova (changes requested):
Though, this has been added to support git mirrors with llvm-zorg. This patch effectively kills this feature.

Let's discuss your use case for "other repourl" before committing this patch shall we?

slydiman:
> Let's discuss your use case for "other repourl"

Case 1: repourl is really a mirror `https://some_mirror/llvm/llvm-project`. Then `GitHubStatusPush._extract_github_info()` will extract the same `repo_owner='llvm'` and `repo_name='llvm-project'`. GitHubStatusPush uses its own API base url, so it will push to `https://github.com/llvm/llvm-project` anyway.

Case 2: repourl is a forked repository `https://github.com/other_owner/other_repo`. This repository may contain commits which are missing in the base repository and vice versa. I don't see any reason to store `https://github.com/llvm/llvm-project` in the DB instead of the correct repourl. 

It seems this patch reverts https://github.com/llvm/llvm-zorg/commit/4bfc47e723ef5e7ba306d09b18fa53afbdd0580b

Some examples/details would be very helpful to understand the problem.

gkistanova:
It seems your Case 2 is a subject to a downstream change to llvm-zorg, as it is not intended to support downstream forks. Looks like a trivial downstream change of `_repourl`, though.

Case 1 is what is broken by this patch. Since the URL is a part of a change info, it actually points to that change in the buildbot UI. For example, see https://lab.llvm.org/buildbot/#/changes. One could clock on a change title and go to the change source. In the case of a mirror repo we want users to go to the original source of the commit, i.e. `https://github.com/llvm/llvm-project` even if the change has been delivered from a mirror repo. This is what that comment right above the changed line is talking about.

slydiman:
Ok, the mirror may use other owner and repo name. Then `repository=self.repourl if self.repourl.startswith('https://github.com') else self._repourl` is the best solution.

gkistanova:
In the case of a mirror we want to always point to the original source of the change as the single point of truth, which is `https://github.com/llvm/llvm-project`.

slydiman:
`if self.repourl.startswith('https://github.com')` it is not a mirror. It is another repository in github.com, which cannot be a mirror. Otherwise `self._repourl` will be used. I can update this patch to 
```
repository=self.repourl if self.repourl.startswith('https://github.com') else self._repourl
```
or we can just close this PR and I will make local changes for test/debug purpose.

gkistanova:
> ```repository=self.repourl if self.repourl.startswith('https://github.com') else self._repourl```

That would work, I think.

It would be nice to support SSH as well, but that could wait till somebody actually needs it.

gkistanova (approved):
LGTM.

PR #112, opened 2024-01-28
[HIP] Add libunwind to LLVM_ENABLE_RUNTIMES

jplehr:
It seems that this is needed now.

Should fix the current state of the clang-hip-builder.
https://lab.llvm.org/buildbot/#/builders/165/builds/48048


jplehr:
Alternatively, we could probably also disable that unwinder setting that the error message mentions.

ronlieb (approved):

jplehr:
I also just realized that the HIP builder is in production, so these changes will likely not propagate automatically. Idk

gkistanova (approved):
LGTM.

PR #115, opened 2024-01-30
Add 2 new mac workers/builders as well as reallocate resources on some existing workers.

dyung:
Adding two new mac builders since I will soon have some spare machines lying around idle and we don't seem to have much mac coverage on the main buildbot.

Also reduced the number of parallel threads used by a few other workers that were sharing another machine's distributed compilation resources (that machine will be redeployed as one of the new mac workers).


gkistanova (approved):
LGTM.

PR #121, opened 2024-02-09
[PPC] Use a symlinked toolchain on the rhel bot

kamaub:
The clang-ppc64le-rhel buildbot uses a different external compiler
toolchain to the system default, this change allows version control
for the selected toolchain without zorg code changes by using a
soft symlink.


buildbot/osuosl/master/config/builders.py
```diff
@@ -689,8 +689,8 @@
                         "-DCMAKE_CXX_COMPILER=clang++",
                         "-DCLANG_DEFAULT_LINKER=lld",
                         "-DLLVM_TOOL_GOLD_BUILD=0",
-                        "-DCMAKE_C_COMPILER_EXTERNAL_TOOLCHAIN:PATH=/usr",
-                        "-DCMAKE_CXX_COMPILER_EXTERNAL_TOOLCHAIN:PATH=/usr",
+                        "-DCMAKE_C_COMPILER_EXTERNAL_TOOLCHAIN:PATH=/gcc-toolchain/usr",
+                        "-DCMAKE_CXX_COMPILER_EXTERNAL_TOOLCHAIN:PATH=/gcc-toolchain/usr",
```
lei137, line 693: where will the actual toolchain used be installed?  I think it's best if we have this path to be setup at the root of where all future toolchains are expected to be installed.

kamaub: > where will the actual toolchain used be installed? 

Redhat toolchains usually go under `/opt/rh/` but that is up to `dnf` installation defaults to decide.
We only need to ensure each instance has a symlink from `/gcc-toolchain/usr` to the desired toolchain, right now that is `/opt/rh/gcc-toolset-12/root/`.

>  I think it's best if we have this path to be setup at the root of where all future toolchains are expected to be installed.

This would result in the same problem Galina pointed out via email, if `/gcc-toolchain/` points to `/opt/rh/` instead of `/opt/rh/gcc-toolset-12/root/` then we will have to set `-DCMAKE_C_COMPILER_EXTERNAL_TOOLCHAIN:PATH=/gcc-toolchain/gcc-toolset-12/root/usr` in zorg code again, making toolchain version control more difficult.

lei137: I was thinking mabye it would be better to link the 2 locations together.  Then it can be seen very clearly that there is a dependency there. ie `/opt/rh/gcc-toolchain->/opt/rh/gcc-toolset-12/root`
Then here we point the toolchain path to `/opt/rh/gcc-toolchain/usr`

lei137 (approved):
LGTM
I'm not tied to it.

kamaub:
FYI: @gkistanova

gkistanova:
Looks like something is not right with your setup of the toolchain. https://lab.llvm.org/buildbot/#/builders/57/builds/32757/steps/4/logs/stdio

Could you move this worker to staging and resolve the issue before returning it back to production, please?

kamaub:
> Looks like something is not right with your setup of the toolchain. https://lab.llvm.org/buildbot/#/builders/57/builds/32757/steps/4/logs/stdio
> 
> Could you move this worker to staging and resolve the issue before returning it back to production, please?

My apologise for the oversight, I have corrected the error in the setup and moved the bot back into production, thank you very much.

PR #126, opened 2024-02-22
[libc] Update libc build config

jplehr:
How to build libc for GPU is about to change significantly, which requires an update to the way it is built in the bot.

After some back and forth, it appears that it cannot be easily added to the `-DLLVM_ENABLE_RUNTIMES` and instead requires some special handling. Thus, for the libc bot we intentionally work around the `enable_runtimes=auto` option and instead enable `'openmp'` and `'compiler-rt'` manually in the bot specification.
In addition, we specify the new CMake flags for the libc build on AMDGPU.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1874,6 +1874,8 @@ all += [
     'factory' : OpenMPBuilder.getOpenMPCMakeBuildFactory(
                         clean=True,
                         depends_on_projects=['llvm', 'clang', 'compiler-rt', 'libc', 'lld', 'openmp'],
+                        # Special case this bot to account for new (verbose) libc build syntax
+                        enable_runtimes=['openmp', 'compiler-rt'],
                         extraCmakeArgs=[
                             "-DCMAKE_BUILD_TYPE=Release",
                             "-DCLANG_DEFAULT_LINKER=lld",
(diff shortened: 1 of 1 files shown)


jhuber6 (approved):

buildbot/osuosl/master/config/builders.py
```diff
@@ -1883,10 +1885,8 @@
                             "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache",
                             "-DLIBOMPTARGET_FOUND_AMDGPU_GPU=ON",
                             "-DLIBOMP_ARCHER_SUPPORT=OFF",
-                            "-DLIBC_GPU_BUILD=ON",
-                            "-DLIBC_GPU_ARCHITECTURES=gfx906",
-                            "-DLIBC_GPU_TEST_ARCHITECTURE=gfx906",
-                            "-DLIBC_GPU_TEST_JOBS=8",
+                            "-DRUNTIMES_amdgcn-amd-amdhsa_LLVM_ENABLE_RUNTIMES=libc",
+                            "-DLLVM_RUNTIME_TARGETS='default,amdgcn-amd-amdhsa'",
```
jhuber6, line 1889: Don't think these are necessary, the Python likely handles the qoutes for us since it does stuff like `"-DLLVM_TARGETS_TO_BUILD=X86;AMDGPU",` above.

jhuber6 (approved):

jplehr:
Hi @gkistanova  I am going to land this as is. Please feel free to comment here if I need to do some fix ups.
I guess, our plan is to migrate the libc bot to an annotated builder at some point, so we went ahead with this somewhat undesirable solution.

PR #128, opened 2024-03-03
Add a new Darwin worker and release builder

dyung:
Add a new release builder for the AArch64 Darwin target, and a new worker to share build responsibilities for the llvm-clang-aarch64-darwin builder.


gkistanova (approved):
LGTM.

I assume you will remove `jobs` from this patch to be inline with https://github.com/llvm/llvm-zorg/pull/130.

dyung:
> I assume you will remove `jobs` from this patch to be inline with #130.

 I don't need to on these workers as they are static and won't be moved around anytime soon.

PR #130, opened 2024-03-05
Make a few buildbot changes to the SIE buildbots

dyung:
I am shuffling around the builder assignments for a few workers, so needed to remove the hard coded values as they were causing the machines to become unresponsive when I scaled down the hardware. Also update the llvm-clang-x86_64-gcc-ubuntu release builder to use the sie-linux-worker3 worker (I forgot to update this in a previous change when I updated the non-release job to use this worker).

1. Remove jobs property for 2 linux workers
2. Switch llvm-clang-x86_64-gcc-ubuntu-release to sie-linux-worker3
3. Remove -j value for some LLVM_LIT_ARGS for 2 builds
4. Remove argument LLVM_PARALLEL_LINK_JOBS for 2 builders


gkistanova (approved):
LGTM.

PR #132, opened 2024-03-08
Revert "Add -DLLVM_DISABLE_ASSEMBLY_FILES=ON to work around a bug in Apple Clang (#129)"

dyung:
Updating to Apple Clang 1500.3.9.4 fixes the problem that was causing test failures.

Build/test is now passing: https://lab.llvm.org/staging/#/builders/188/builds/444.


gkistanova (approved):
LGTM.

PR #133, opened 2024-03-11
Try adding an option to automatically install MLIR pip requirements to getCmakeWithNinjaBuildFactory

joker-eph:
Trying to fix the pre-merge bots which are broken right now.

I can't test it though since I don't have SSH access to these bots.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -3003,6 +3003,7 @@ all += [
                     vs="autodetect",
                     depends_on_projects=["clang-tools-extra", "clang", "flang", "libclc", "lld", "llvm", "mlir", "polly", "pstl"],
                     checks=["check-all"],
+                    install_pip_requirements = True,
                     extra_configure_args=[
                         "-DCMAKE_BUILD_TYPE=Release",
                         "-DLLVM_ENABLE_ASSERTIONS=ON",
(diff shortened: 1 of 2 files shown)


joker-eph:
1) We should do the same for the premerge config, I'll follow-up on this
2) It'd be great if @lnihlen or @metaflow could try to run `python3 -m pip install -r mlir/python/requirements.txt` on the linux and the windows bot and confirm if this invocation is working there?

joker-eph:
Ping @gkistanova @lnihlen @metaflow : the bots are broken until we resolve this!

metaflow:
> Trying to fix the pre-merge bots which are broken right now.

I am fine with merging this (again: I have not tested it) but maybe it's a better idea to roll back the breaking change and test it first?

>I can't test it though since I don't have SSH access to these bots.

You can try to run build in docker https://github.com/google/llvm-premerge-checks/blob/main/containers/buildbot-linux/Dockerfile that is used for those buildbots.

metaflow (approved):

joker-eph:
@metaflow do you know why the behavior is different in the buildkite environment than the buildbot? Isn't the docker image the same?

Also I won't be able to run the windows container: Docker requires a windows host for this I believe.

joker-eph:
Actually didn't mean to merge without @gkistanova's review: this is touching the  buildbot factory logic, can't reopen a merged reverted PR on GitHub, I think I have to open a new one...

PR #136, opened 2024-03-13
Try adding an option to automatically install MLIR pip requirements to getCmakeWithNinjaBuildFactory

joker-eph:
Trying to fix the pre-merge bots which are broken right now.

This adds the `pip install` command that is included in the pre-merge config:

https://github.com/llvm/llvm-project/blob/0a443f13b49b3f392461a0bb60b0146cfc4607c7/.ci/monolithic-windows.sh#L40


gkistanova (approved):
This should solve the immediate problem. Later I'll think if there is a way to generalize requirements.

joker-eph:
That worked! :)

PR #137, opened 2024-03-14
[flang] Enable OpenMP project for Linaro's bots

luporl:
As discussed in Discourse and proposed in
https://github.com/llvm/llvm-project/pull/80874, Flang will start
to use C/C++ header files and Fortran modules from OpenMP runtime.

When the mentioned PR lands, all of Flang's OpenMP tests will be
skipped as unsupported, unless the OpenMP project is enabled.


gkistanova (approved):
LGTM.

luporl:
@gkistanova thanks for the review!

PR #138, opened 2024-03-18
Fix hip-build.sh to build llvm/runtime together

yxsamliu:
Previsouly the bot build runtime first using the old llvm/clang.

This does not work of libc++ or other runtime requires new llvm/clang.


jplehr (approved):
LGTM

PR #141, opened 2024-03-19
[OpenMP/Offload] Enable Offload runtime and add check

jplehr:
This enables the new "offload" runtime that
https://github.com/llvm/llvm-project/pull/75125 will create. In addition, it adds a new lit test target.
I removed the extra CMake argument that would again specify which runtimes to build manually, so that we rely on the factory to add the respective runtimes.

Since the PR does not make `check-libomptarget` available in the top level configuration, we have to change the ninja working directory into the runtimes build folder, where the target is known. Whether or not this works with this crude "additional lit checks" mechanism is an experiment.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1832,14 +1832,13 @@ all += [
     'builddir': "openmp-offload-amdgpu-runtime-2",
     'factory' : OpenMPBuilder.getOpenMPCMakeBuildFactory(
                         clean=True,
-                        enable_runtimes=['openmp'],
+                        enable_runtimes=['openmp','offload'],
                         depends_on_projects=['llvm','clang','lld','openmp'],
                         extraCmakeArgs=[
                             "-DCMAKE_BUILD_TYPE=Release",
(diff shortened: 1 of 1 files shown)


jplehr:
A couple of more comments:
- This affects: https://lab.llvm.org/staging/#/builders/185
- I think we should wait for @saiislam to let us know of the status on his downstreaming efforts of the llvm-project PR before proceeding with it and use this buildbot change as a test in the meantime to make sure that the `add_lit_check` works as intended.
- This will make the bot red if it works. The message should be that it `cd`ed into the directory, but the target name is unknown. If the bot prints a different error, adding this target via the `add_lit_checks` liekly did not work.

jplehr:
I just realized that I'll have to change the patch to test the `add_lit_checks` mechanism. As it is now, the bot will bail out early, given that `offload` is unknown.

jplehr:
I created https://github.com/llvm/llvm-zorg/pull/142 which should only add the additional lit target. Meaning that we can use it to see if the machinery is able to handle what we need.

jplehr:
I'll close this PR, since we discussed a different route during a meeting today.

PR #142, opened 2024-03-19
[OpenMP/Offload] Experiment w/ additional lit target

jplehr:
This commit adds a lit target that has to change the ninja workdir for a target to be picked-up. This is meant as a test as  I'm unsure whether the somewhat crude implementation of add_lit_check is capable to handle it.


jplehr:
I think we settled on a different route during our meeting today.

PR #153, opened 2024-03-26
[green-dragon-migration] Update task shell scripts

justice-adams-apple:
## In This PR

* Update task shell scripts to use python 3
* Update any shell scripts pulling the compiler to install the awscli dependency needed
* Don't require virtualenv as with python3 we just use venv built into the python stdlib


justice-adams-apple:
cc @azharudd

azharudd (approved):

PR #155, opened 2024-03-26
[green-dragon-migration] delete build.py

justice-adams-apple:
## In This PR
* Delete build.py

Build.py is an old build script from the ages of before the monorepo. It is not used in any CI jobs. We should delete it as it can be confusing for contributors looking to update the build script, See this PR which actually has no effect: https://github.com/llvm/llvm-zorg/pull/144


justice-adams-apple:
cc @azharudd

azharudd:
I was thinking of keeping this around for old history, while renaming it to something like `old_build.py` or `splitrepo_build.py`. We can rename `monorepo_build.py` to `build.py` and update all its invocations to that, while keeping a symlink `monorepo_build.py` -> `build.py` just in case for now. What do you think?

justice-adams-apple:
@azharudd https://github.com/llvm/llvm-zorg/pull/156

PR #156, opened 2024-03-29
[green-dragon-migration] rename build scripts

justice-adams-apple:
Follow up from https://github.com/llvm/llvm-zorg/pull/155

* rename build.py -> old_build.py (and associated test file)
* rename monorepo_build.py -> build.py (and associated test file)
* create symlink from monorepo_build.py -> build.py


azharudd:
Thanks!

PR #159, opened 2024-03-29
rewrite relay to not use insecure runtime method signatures and pull …

justice-adams-apple:
## In This PR
Rewrite relay.groovy, namely the following changes were made:
* Take in a list of jobs rather than a pattern. We do this so that we don't have to rely on `Jenkins.getInstance()`, `job.getName()`, or `jenkins.getAllItems()` which are insecure Jenkins methods and will result in security vulnerabilities if we approve them.
* Pull artifacts from S3 storage
* Implement a fix for LNT server overloading until we figure out issues with the server

### DIFF
--- a/zorg/jenkins/relay.groovy
+++ b/zorg/jenkins/relay.groovy
@@ -1,43 +1,31 @@
 #!/usr/bin/env groovy
-@NonCPS
-private def get_matching_jobs(pattern) {
-    def jobs = []
-    for (job in Jenkins.getInstance().getAllItems(Job)) {
-        def jobname = job.getName()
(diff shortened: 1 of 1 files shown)


justice-adams-apple:
A relay job would now look like this:
```
jobs = [
    "lnt-ctmark-aarch64-O0-g",
    "lnt-ctmark-aarch64-O3-flto",
    "lnt-ctmark-aarch64-Os",
    "lnt-ctmark-aarch64-Oz",
]

relay.pipeline jobs
```

azharudd (approved):
LGTM. Thanks!

PR #161, opened 2024-04-01
[green-dragon-migration] Update common.groovy

justice-adams-apple:
## In This PR
* Update coming.groovy to use build URL for consoleText
* Pull from S3
* Don't send notification emails


justice-adams-apple:
cc @azharudd

PR #162, opened 2024-04-01
[green-dragon-migration] Add stage 1 jobs

justice-adams-apple:
## In This PR
* Update all existing stage 1 jobs
* Add declarative pipeline for stage1 expensive checks

All stage1 jobs have been updated to:
* Run on the correct labels
* upload artifacts to S3
* Set environment variables in `environment` closure when possible
* Setup python3 venv with all dependencies to build/test

### DIFF
--- a/zorg/jenkins/jobs/jobs/clang-san-iossim
+++ b/zorg/jenkins/jobs/jobs/clang-san-iossim
@@ -1,61 +1,88 @@
 pipeline {
-    agent { label 'green-dragon-08' }
-
     parameters {
-        string(name: 'GIT_REVISION', defaultValue: '*/main', description: 'Git revision to build')
+        string(name: 'LABEL', defaultValue: params.LABEL ?: 'macos-x86_64', description: 'Node label to run on')
+
+        string(name: 'GIT_REVISION', defaultValue: params.GIT_REVISION ?: '*/main', description: 'Git revision to build')
+    }
+
+    agent {
+        node {
+            label params.LABEL
+        }
     }
 
     stages {
         stage('Checkout') {
             steps {
                 dir('llvm-project') {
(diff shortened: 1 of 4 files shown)


justice-adams-apple:
cc @azharudd

zorg/jenkins/jobs/jobs/clang-san-iossim
```diff
@@ -78,26 +105,23 @@ pipeline {
                       $COMPILER_RT_TEST_DIR/ubsan/ThreadSanitizer-iossim-x86_64/
                     '''
                 }
-                junit 'clang-build/**/testresults-*.xunit.xml'
+            }
+            post {
+                always {
+                    script {
+                        junit allowEmptyResults: true, testResults: 'clang-build/**/testresults-*.xunit.xml'
```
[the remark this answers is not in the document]
justice-adams-apple: Updated to remove the allowEmptyResults

zorg/jenkins/jobs/jobs/clang-stage1-RA
```diff
@@ -1,6 +1,10 @@
 pipeline {
+    options {
```
azharudd, line 2: We should let these do concurrent builds, but throttle on the number of concurrent builds it can do. That would provide a nice balance and keep the changelist short while not using too many nodes.

azharudd (commented):
One more suggestion, LGTM otherwise.

zorg/jenkins/jobs/jobs/clang-stage1-RA
justice-adams-apple: @azharudd I think this would require us to add the [Throttling concurrent builds plugin](https://plugins.jenkins.io/throttle-concurrents/) I believe, so I would propose we do this in a later PR once everything is established. Right now we handle ~10-20 commits on average in our stage1 jobs

PR #163, opened 2024-04-02
[GreenDragon] Workaround Stage 2 bots being broken

justice-adams-apple:
Simply mimicking https://github.com/llvm/llvm-zorg/pull/144/

The change was made to the wrong build script initially


justice-adams-apple:
cc @azharudd

PR #164, opened 2024-04-03
[flang] Disable OpenMP on flang-aarch64-libcxx bot

luporl:
Building LLVM with libc++ and OpenMP as a runtime on a glibc based
system causes some hard to fix build failures (see
https://github.com/llvm/llvm-project/issues/61693).

As the point of building against libc++ was to make sure flang works
well with it, adding OpenMP doesn’t help that, and flang OpenMP
tests are already enabled on other flang bots, just disable OpenMP
on flang-aarch64-libcxx.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2242,7 +2242,7 @@ all += [
     'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
                     clean=True,
                     checks=['check-flang'],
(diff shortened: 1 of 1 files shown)


DavidSpickett:
I agree with the change as is, is it also worth writing down somewhere any alternatives you tried? If it's not possible at all right now, perhaps as an issue so anyone who does require this build config can work from that. Then link to it from this commit's message.

luporl:
Thanks for the review. I've added a link to an existing issue, that explains the problem in more depth and suggests a workaround.

DavidSpickett (approved):
We have an audit trail for the decision, so this LGTM.

PR #165, opened 2024-04-05
Enable FLANG_LIBCUDACXX_PATH for 'flang-runtime-cuda-gcc' builder.

vvereschaka:
Pass a path to LIBCUDACXX source code to avoid the following build warnings:

  warning #20014-D: calling a __host__ function ...


vzakhari (approved):
Thank you, Vlad!

gkistanova (approved):
LGTM.

PR #174, opened 2024-04-15
Use updated lunch configuration format

hiraditya:
With this patch lunch target will be aosp_arm64-trunk-userdebug
Doc:  https://source.android.com/docs/setup/build/building#view_the_current_target

Should fix buildbot failure: https://lab.llvm.org/buildbot/#/builders/82/builds/15143


stephenhines (approved):
This looks fine to me, but please double check that they've updated to a recent enough AOSP to require the `trunk` argument to the lunch command line.

PR #175, opened 2024-04-16
[OpenMP][AMDGPU] Build compiler-rt as runtime

jplehr:
This adds compiler-rt as runtime to build to unify our buildbots and removes the unneeded manual addition of openmp as a runtime to build in the CMake options list.


Meinersbur (approved):
LGTM

PR #176, opened 2024-04-22
[OpenMP] Add offload to other buildbots

jplehr:
This is in anticipation of landing the move to offload. We have one buildbot that already runs the configuration with offload enabled, which will hopefully turn green once the patch lands.
Once things are fine on that end, we can land this patch to enable the offload runtime in the other buildbots as well.


saiislam (approved):
The changes look good to me. But, please wait for someone who is more knowledgeable in bot management to review it.

jplehr:
Thanks Saiyed. I will wait for this patch to be landed until the staging offload bot turned green -- so that we know what is a working configuration.

PR #177, opened 2024-04-23
[Offload] Make buildbots depend on offload project

jplehr:
This is making the AMDGPU offload buildbots depend on the 'offload' project. FWICT this should work with the current buildbot implementation we have, but someone more knowledable should confirm that this works.


Meinersbur (approved):
Should work. `depends_on_projects` determines changes in which top-level directories trigger a build.  By default it is `depends_on_projects = ['llvm', 'clang', 'openmp']`.

PR #182, opened 2024-04-26
[BOLT] Drop BOLT_CLANG_EXE

aaupov:
As a follow-up to the issue unearthed by
https://github.com/llvm/llvm-project/pull/89693, remove all uses of
BOLT_CLANG_EXE to prevent similar blockers in the future and reduce builder
maintenance overhead.


MaskRay (approved):
Thanks!

I guess build bots use `ninja all`. If adding `clang` builds too many executables, perhaps explicit `ninja clang` will help.

aaupov:
> Thanks!
> 
> I guess build bots use `ninja all`. If adding `clang` builds too many executables, perhaps explicit `ninja clang` will help.

No, BOLT buildbots are configured to only run `ninja check-bolt`.

PR #183, opened 2024-04-30
[flang] Enable FLANG_ENABLE_WERROR and LLVM_ENABLE_ASSERTIONS in ppc64le-flang build

kkwli:


DanielCChen (approved):
The change makes it consistent with the build script provided for flang on github. Thanks for fixing it.

PR #188, opened 2024-05-09
[RISCV] Add new "native" (qemu-system) builders

asb:
For background, I don't expect these builders to graduate from the staging buildmaster in the current configuration. I'm spinning them up so we get some slow feedback while I, in parallel, explore alternative build configs that make better tradeoffs in terms of response time vs feedback.

These are all qemu-system based. Please note that the first commit adds a new ability to set cmake flags that apply only to the second-stage build, which I use to apply `-march` that the system compiler may not understand as well as to have ccache enabled for the first stage only. Happy to split that out to a separate PR if you prefer.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -3126,6 +3126,95 @@ all += [
                         'CXX': 'clang++',
                     })},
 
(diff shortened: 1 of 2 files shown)


gkistanova (approved):
LGTM.

asb:
Thanks for the review, I've emailed you the builder passes.

PR #189, opened 2024-05-13
Clean source dir on request in PollyBuilder.

efriedma-quic:


Meinersbur:
"On request" is `step.build.getProperty("clean")`. Using the `clean` argument will clean the source directory every single build for builders where this is set (e.g. `polly-arm-linux`) and redownload >1GB from GitHub. I think we should not abuse the available bandwidth.

efriedma-quic:
The reason this came up was that there was a bad commit in LLVM that stuck some files into the source tree, so then the buildbot was continuously failing.  We ended up logging into the buildbot to fix it by hand.

Is there some way to force a clean through the buildbot interface?  Not sure how you do that, or what kind of permissions are required.

Can we make the buildbot run "git clean" to fix this sort of issue automatically, without re-downloading the whole tree?

Meinersbur:
When you login and your GitHub email matches the email in the zorg config, a couple of options become available:

image

image

The `step.build.getProperty("clean")` property is set using this flag: 

image

Both flags together don't always work as expected for all factories but I've been running against walls with that one already.

Local git repositories can fail a lot of different ways (e.g. disk corruption) and there is no one-solution-fits-them-all to fix it. I think occasionally re-cloning a broken repository manually is fine. Would be nice if buildbot would itself detect a broken repository but that would be an upstream feature.

efriedma-quic:
I suspect a "git clean" handles the most common cases, so maybe worth doing even if it doesn't fix everything?

Otherwise, I guess if this is working as intended, we can leave it, sure.

Meinersbur:
Buildbot does `git reset --hard` to checkout the next revision. What would `git clean` fix that `git reset --hard` doesn't?

efriedma-quic:
`git reset --hard` only affects tracked files.  git clean would delete untracked files (which lit tests can introduce by accident pretty easily).

gkistanova (commented):
This patch would change the expected behavior (i.e. how it works for other builders). `clean` argument when True means "clean build directory", not the source code as well.

Pulling the source code from scratch for every build is not a good idea. Unless I'm missing something, forcing a clean checkout from WebUI on a special occasion covers the described case.

Meinersbur:
@efriedma-quic Buildbot seems to support what we want according to the documentation: https://docs.buildbot.net/current/manual/configuration/steps/source_git.html#step-Git

 * `clobberOnFailure=True`: If a fetch or full clone fails, we can retry to checkout the source by removing everything and cloning the repository. If the retry fails, it fails the source checkout step.
 * `mode=full`: The build tree is clean of any built files. The exact method for doing this is controlled by the method argument.
 * `method=fresh`: This removes all other files except those tracked by Git. First it does `git clean -d -f -f -x`, then fetch/checkout to a specified revision (if any). This option is equal to update mode with `ignore_ignores=True` in old steps.

PR #191, opened 2024-05-15
[Offload] Add RHEL 8.8 based buildbot

jplehr:


jhuber6 (approved):

buildbot/osuosl/master/config/builders.py
```diff
@@ -2019,6 +2019,38 @@
...
 
+    {'name' : "openmp-offload-rhel-8.8",
+    'tags'  : ["openmp"],
+    'workernames' : ["rocm-worker-hw-04-rhel-8.8"],
+    'builddir': "openmp-offload-rhel-8.8-build",
+    'factory' : OpenMPBuilder.getOpenMPCMakeBuildFactory(
+                        clean=True,
+                        test=True,
+                        enable_runtimes=['openmp', 'compiler-rt', 'offload'],
+                        depends_on_projects=['llvm','clang', 'flang', 'lld', 'mlir', 'offload', 'openmp'],
```
gkistanova, line 2030: This builder builds compiler-rt, but it does not listen for compiler-rt commits. This leads to false blames when a compiler-rt commit would break a build but it will be detected only on unrelated commits to the projects the builder depends on. This is not good.

Any reason not to have compiler-rt in the `depends_on_projects` list?

gkistanova (commented):
The patch looks Ok in general. One thing, though.

buildbot/osuosl/master/config/builders.py
```diff
@@ -2019,6 +2019,38 @@
                         add_openmp_lit_args=["--time-tests", "--timeout 100"],
                     )},
 
+    {'name' : "openmp-offload-rhel-8.8",
+    'tags'  : ["openmp"],
+    'workernames' : ["rocm-worker-hw-04-rhel-8.8"],
```
gkistanova, line 2024: Unfortunately, a worker name must be an identifier, and this one is not. You need to come up with something not containing the dot symbol.

gkistanova (changes requested):

jplehr:
Thank you Galina.
I updated the worker name in this PR and added `compiler-rt` to the `depends_on_projects` list.

buildbot/osuosl/master/config/builders.py
jplehr: Done.

buildbot/osuosl/master/config/builders.py
jplehr: Done

gkistanova (approved):
LGTM.

PR #192, opened 2024-05-16
[Offload][OpenMP] Various fixes to AMDGPU offload bots

jplehr:
This addresses several problems with Offload/OpenMP AMDGPU buildbots:
- Change worker name for RHEL 9.4 to not contain a '.'
- List 'compiler-rt' in 'depends_on_projects' lists to not generate false blame emails as pointed out by Galina in a separate review.
- Update the SLES builder environment to add /opt/rocm/lib path to the LD_LIBRARY_PATH environment variable.


gkistanova (approved):
LGTM.

PR #195, opened 2024-05-20
Disable PDB generation in stage1 of the clang-cl self-host bot

rnk:
PDB generation is resource-intensive and does not perform well for statically linked unit tests. The stage2 build generates PDBs to exercise clang-cl's debug info generation abilities, but we generally do not use the debug info from the stage 1 builds. PDB build errors are also a common source of builder unreliability and lead to excessive disk usage, so disable them for now.

Here is a recent example of PDB unreliability:
https://lab.llvm.org/buildbot/#/builders/123/builds/26898
```
... GeneratorTest.cpp: fatal error C1051: program database file, '...\vc140.pdb', has an obsolete format, delete it and recompile
```

### DIFF
--- a/zorg/buildbot/builders/annotated/clang-windows.py
+++ b/zorg/buildbot/builders/annotated/clang-windows.py
@@ -40,7 +40,7 @@ def main(argv):
                       stages=stages,
                       check_targets=check_targets,
                       extra_cmake_args=extra_cmake_args,
-                      stage1_extra_cmake_args=extra_cmake_args,
+                      stage1_extra_cmake_args=stage1_extra_cmake_args,
                       check_stages=check_stages,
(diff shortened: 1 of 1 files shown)


lnihlen (approved):

rnk:
Thanks! Theoretically, the annotated buildbot recipe will let me submit this and immediately see if my results are reflected on the next build.

lnihlen:
Sounds good. I'm off today, but hobby hacking, so I'm around if you need me to poke something on the cloud side.

rnk:
The stage1 CMake args took effect in this build:
https://lab.llvm.org/buildbot/#/builders/123/builds/26903

It will take two hours to finish the self-host build, but I expect it will come back green this time.

PR #196, opened 2024-05-28
Fix ClangBuilder after f59269f1ef7823fe68775d9a2fcc975673654bae

DavidSpickett:
The way that change was indented meant that the cmake/build/check steps of stage 2 were only added if extra_stage2_cmake_args was not empty.

This caused any 2 stage builder without those extra options to skip building stage 2. Which left them green but not doing their
job, or red if they tried to run the test suite using stage 2.

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -436,17 +436,17 @@ def _getClangCMakeBuildFactory(
                             stage1_install,
                             cxx))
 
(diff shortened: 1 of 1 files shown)


antmox (approved):
looks good to me

asb (approved):
Thanks for tracking that down, and sorry for the breakage. Is there a lint check that would have flagged this as a potential error?

DavidSpickett:
Maybe static analysis could have seen that stage1_cc is undefined if useTwoStage is False, but was used without checking that useTwoStage is True.

PR #203, opened 2024-06-10
sparc64 linker and jobs changes

glaubitz:
Some changes to the clang-sparc64-linux configuration in the hope to fix the linking timeouts.


gkistanova (approved):
LGTM.

PR #204, opened 2024-06-13
Fix 2-stage builders

luporl:
After the LLVM buildbot maintenance, 2-stage builders started to
fail with the following error:
  Unknown value for CMAKE_BUILD_TYPE: {stage2_config}


luporl:
Note that this was not tested in a local environment.

antmox (approved):
LGTM

fhahn:
Thanks for the fix! Could we get this merged soon to unblock failing builders?

luporl:
Merged. Thanks for the reviews!

PR #205, opened 2024-06-14
ClangBuilder: fix cmake error due to wrong f-string

antmox:
AArch64 stage-2 bots were failing with:
  Unknown value for CMAKE_BUILD_TYPE: {stage2_config}

Formatting was not done due to missing 'f' prefix

Issue reported by F.Hahn

Introduced by 8127be09b18be742911a01fef5d9d5f93d5ae329


antmox:
oops a pr already exists #204

PR #208, opened 2024-06-19
Updated the regex to extract PR# from the PR title

slydiman:


gkistanova:
Could you show the case you are trying to handle, please? A reference to the actual commits to llvm-project is fine.

PR #209, opened 2024-06-19
Use lower priority for 'preamble' logs. Use higher priority for the rest stdio.

slydiman:
Here is the sample https://lab.llvm.org/buildbot/#/builders/101/builds/263

The failed step 3 has 2 logs `preamble` and `stdio` with the same type `s`. We must select the second log in this case.


gkistanova (approved):
LGTM.

PR #210, opened 2024-06-19
Added the summary to the github comment

slydiman:
The timeout error description (`command timed out: 1200 seconds without output running ..., attempting to kill`) in LitTestCommand can be retrieved only via getResultSummary(). Without the summary the log is useless in this case.


gkistanova:
Could you add to the description a reference to a particular build results and show the text before and after this change, please? As well as a "normal" case without the timeout. This information will be helpful.

PR #212, opened 2024-06-20
[OpenMP][Offload] Enable check-offload for CUDA OpenMP builders

Meinersbur:
In contrast to `check-openmp`, `check-offload` is not implicit. Also add `--time-tests` to be able to see the duration of each test.

Thanks to @jplehr for the hint.


gkistanova (approved):
LGTM.

PR #213, opened 2024-06-21
Replace deprecated CUDA_TOOLKIT_ROOT_DIR with CUDAToolkit_ROOT

Meinersbur:
LLVM does not use CUDA_TOOLKIT_ROOT_DIR anymore after https://github.com/llvm/llvm-project/commit/3de162fee44376b3857c73958d656f76b17e50f5. Instead use `CUDAToolkit_ROOT` to let CMake search for CUDA.


jhuber6 (approved):
Thanks, not sure this was ever correct in the first place since AFAIK it was just an empty string until the linked patch.

Meinersbur:
It did work before the liboffload refactor (buildbot was green at some point), but I don't know how `CUDA_TOOLKIT_ROOT_DIR` was forwarded into the runtimes-bin directory.

PR #217, opened 2024-06-28
[OpenMP][AMDGPU] Disable archer tests

jplehr:
There appears to be one particular libarcher test that is quite flaky on the dockerized RHEL bots. Disable libarcher all together to reduce noise in testing.


ronlieb (approved):

gkistanova (approved):
LGTM.

PR #222, opened 2024-07-09
[zorg] Gently pass through the IRenderables in factory.pathRelativeTo static method.

vvereschaka:
We can get `IRenderable` (util.Interpolate) into `factory.pathRelativeTo()` from the builder configuration. We cannot process IRenderables here because of lack of required data. Just gently pass through that kind of arguments.


gkistanova (approved):
LGTM.

PR #223, opened 2024-07-10
Add a new 'hint' argument for LLVMBuildFactory factory class.

vvereschaka:
This parameter is used to apply suffixes to the step names when factory is used as a nested factory for another one. The suffix will be added to the step name separated by dash symbol.

As example for 'stageX' hint the following step names will be generated for the factory's steps:
    cmake-cofigure => cmake-configure-stageX
    build => build-<target>-stageX


vvereschaka:
image
This is an example of the builder's steps for nested factory with the hint argument as `aarch64-lldb`.

gkistanova (approved):
LGTM.

PR #224, opened 2024-07-10
Add new arguments for UnifiedBuildTree.getCmakeExBuildFactory() factory.

vvereschaka:
* repo_profiles - pass a repository profile name ("default") to retrieve the source code or pass None to disable checking out from the repos.
* hint - factory hint.

Both of those new arguments are useful for the nested factories.

Also removed usage of IRenderable/Interpolate on the arguments inside of the factory. The arguments can be IRenderable and we must not wrapping them in IRenderable again.

### DIFF
--- a/zorg/buildbot/builders/UnifiedTreeBuilder.py
+++ b/zorg/buildbot/builders/UnifiedTreeBuilder.py
@@ -608,6 +608,7 @@ def getCmakeExBuildFactory(
         vs = None,                      # VS tools environment variable if using MSVC.
         vs_arch = None,
         clean = False,                  # Do clean build flag.
+        repo_profiles = "default",      # The source code repository profiles.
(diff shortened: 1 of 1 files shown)


vvereschaka:
depends on #223

gkistanova (changes requested):
I see 3 unrelated changes in this PR. Could you split them to 3 separate PRs, please?

One for the "hint", another one for "repo_profiles" (why that's plural, by the way?), and the third one for deleting the "Remove the source code" step.

Thanks!

vvereschaka:
It could be difficult to split these changes for 3 commits. The changes are too close. Is it really necessary?

>another one for "repo_profiles" (why that's plural, by the way?)

 I'm thinking may be in future it would be simpler to switch to multiply profiles to allow builder getting the source code (or something else) from multiply sources. We can do it without changing the argument name.

vvereschaka:
>Could you split them to 3 separate PRs, please?

ok

PR #225, opened 2024-07-11
fix hip-build.sh about compiler-rt

yxsamliu:
Build compiler-rt as runtime instead of project. This will use freshly built clang to build compiler-rt instead of the system compiler. Also use lld to link compiler-rt. Both can lower the risk of encounter issues with old system compiler and linker.

Fixes buildbot failure https://lab.llvm.org/buildbot/#/builders/123/builds/1580


yxsamliu:
Since it is urgent. I will merge it first.

rampitec (commented):
LGTM, thanks! And I didn't know this was an option.

PR #227, opened 2024-07-12
Add new 'repo_profiles' argument for UnifiedTreeBuilder.getCmakeExBuildFactory() factory.

vvereschaka:
A name of the source code profile to get from the remote repository. Currently is supported only "default" profile and None. Default is "default".

If None is passed, no the repository checkout steps will be added to the factory workflow. This is useful for the nested factories when the single repo is shared between them.

A supporting of multiply profiles will be added when it gets supported by Zorg core.


gkistanova (approved):
LGTM.

PR #228, opened 2024-07-12
Add new 'hint' argument for UnifiedTreeBuilder.getCmakeExBuildFactory() factory.

vvereschaka:
This argument can be used to apply suffixes to the step names when factory is used as a nested factory for another one.

The suffix will be added to the step name separated by dash symbol.


gkistanova (approved):
LGTM.

PR #231, opened 2024-07-17
[RISCV] Try to fix linker error for rv32

mikhailramalho:
Currently the buildbot is failing to find lld, so I added a few flags I use locally to force cmake to use lld.


asb (approved):
LGTM

PR #236, opened 2024-07-18
[RISCV] Limit the number the parallel processes for rv32

mikhailramalho:
This patch is intended to limit the number of parallel processes to 8 in the rv32 buildbot, since it's running the tests are run inside qemu and too many processes ends up causing test failures.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2246,7 +2246,7 @@ all += [
     'factory' : AnnotatedBuilder.getAnnotatedBuildFactory(
                     script="libc-linux.py",
                     depends_on_projects=['llvm', 'libc'],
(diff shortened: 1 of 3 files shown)


asb:
By my reading this limits the number of jobs for both the cross-compilation on the host, and the running of tests via the cross.sh helper script. Is there a reasonable way to limit it just for the latter?

mikhailramalho:
> By my reading this limits the number of jobs for both the cross-compilation on the host, and the running of tests via the cross.sh helper script. Is there a reasonable way to limit it just for the latter?

Good idea, there is a separate step to build libc, it can be changed to use all processors

mikhailramalho:
We changed qemu to use shared dirs, so the crash no longer happens when running multiple parallel tests

Issue #235, opened 2024-07-18
test/buildbot/builders/unified_cmakeex.py fails at least since 'hint' was added


asb:
I had a quick look at fixing this but I wasn't completely sure what the current `partly_rendered` logic is trying to achieve.

This is the error, which seems to stem from the format string being passed to `util.Interpolate` now using syntax that isn't really compatible with Python `%` based formatting.

```
Traceback (most recent call last):
  File "/home/asb/repos/llvm-zorg/test/buildbot/builders/unified_cmakeex.py", line 93, in <module>
    assert factory_has_step(f, "Checkout the source code")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/asb/repos/llvm-zorg/test/buildbot/builders/unified_cmakeex.py", line 38, in factory_has_step
    v = partly_rendered(s.kwargs.get("name"))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/asb/repos/llvm-zorg/test/buildbot/builders/unified_cmakeex.py", line 19, in partly_rendered
    return r.fmtstring % interpolations
           ~~~~~~~~~~~~^~~~~~~~~~~~~~~~
KeyError: 'kw:hint:-'
```

vvereschaka:
Thank you @asb ,
I'll update the tests properly.

vvereschaka:
Fixed by #237

PR #237, opened 2024-07-19
Fix failed test/buildbot/builders/unified_cmakeex.py test (issue #235).

vvereschaka:
The changes for UnifiedTreeBuilder.getCmakeExBuildFactory factory and its test:

* updated the step formatting with 'hint' argument (avoid usage of substitutions).
* allowed only 'str' or None for the 'hint' argument.
* updated 'test/buildbot/builders/unified_cmakeex.py' test for the factory changes accordingly.

See #235 for more details.

### DIFF
--- a/test/buildbot/builders/unified_cmakeex.py
+++ b/test/buildbot/builders/unified_cmakeex.py
@@ -11,11 +11,16 @@ import zorg
 from zorg.buildbot.builders import UnifiedTreeBuilder
 from zorg.buildbot.process.factory import LLVMBuildFactory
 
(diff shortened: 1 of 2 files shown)


asb (approved):
LGTM in that the test now passes, and to someone not deeply familiar with the code nothing stands out as questionable. Especially given the status quo is that the test doesn't work at all I think this LGTM is probably sufficient, but if you have doubts about and your changes best to wait for additional review.

gkistanova (approved):
LGTM.

PR #234, opened 2024-07-18
[test] Fix errors in Import.py test

asb:
This test had Python2 style prints, and needs to call ClangBuilder.getClangCMakeBuildFactory rather than ClangBuilder.getClangBuildFactory.

PR #238, opened 2024-07-19
Fix test/buildbot/builders/Import.py test accordingly the current Zorg version.

vvereschaka:
Update accordingly the current version of Zorg's buildbot factories and supported min python version (3.6).


gkistanova:
This looks like a duplicate of this patch - https://github.com/llvm/llvm-zorg/pull/234/files. Alex seems got it first.

vvereschaka:
Yes, right. I didn't notice Alex's patch, sorry. I'll close this PR in that case.

PR #243, opened 2024-07-24
[libc] Rename riscv32 buildbot to enable fullbuild

mikhailramalho:
The libc buildbots use the builder's name to enable/disable libc features, so we add 'fullbuild' to the bot's name to enable libc's fullbuild.


asb (approved):
LGTM.

Is it worth removing "-debian" from the name too, given this is actually a yocto image?

mikhailramalho:
> LGTM.
> 
> Is it worth removing "-debian" from the name too, given this is actually a yocto image?

Indeed, I'll replace it

PR #244, opened 2024-07-25
Update 'lldb-remote-linux-ubuntu' builder, remove --min/max-gdbserver-port arguments.

vvereschaka:
Removed `--min/max-gdbserver-port` arguments when executing the lldb-server instance on the remote target host. Looks like these arguments are not working properly, lets just avoid using them for awhile.

Removed '--log-channel' and '--log-file' arguments for the lldb-server instance as useless for this run configuration.

Removed the remote target rebooting after the tests get done.


vvereschaka:
I'm going to remove the --log-file argument for lldb-server.

vvereschaka:
Also temporary removed the remote target reboot after the tests.

andreil99 (approved):
Thanks, Vlad. Looks good.

PR #245, opened 2024-07-28
Support testsuite builds without LNT

omjavaid:
This adds support in Clang builder to run LLVM testsuite without LNT. This will help us enable LLVM WoA testsuite bot and also help us run LLVM testsuite without LNT on Linux.

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -154,7 +154,7 @@ def getClangCMakeBuildFactory(
             testsuite_flags=None,
             submitURL=None,
             testerName=None,
-
+            testWithLNT=True,
             # Environmental variables for all steps.
             env=None,
             extra_cmake_args=None,
@@ -176,7 +176,7 @@ def getClangCMakeBuildFactory(
                vs_target_arch=vs_target_arch, useTwoStage=useTwoStage,
                testStage1=testStage1, stage1_config=stage1_config,
                stage2_config=stage2_config, runTestSuite=runTestSuite,
-               nt_flags=nt_flags, testsuite_flags=testsuite_flags,
+               nt_flags=nt_flags, testsuite_flags=testsuite_flags, testWithLNT=testWithLNT,
                submitURL=submitURL, testerName=testerName,
                env=env, extra_cmake_args=extra_cmake_args,
                extra_stage2_cmake_args=extra_stage2_cmake_args,
@@ -209,6 +209,7 @@ def _getClangCMakeBuildFactory(
             runTestSuite=False,
             nt_flags=None,
             testsuite_flags=None,
+            testWithLNT=True,
             submitURL=None,
             testerName=None,
 
@@ -284,10 +285,12 @@ def _getClangCMakeBuildFactory(
     # and the test-suite separately. Le's do this first,
     # so we wouldn't poison got_revision property.
     if runTestSuite or checkout_test_suite:
-        f.addGetSourcecodeForProject(
-            project='lnt',
-            src_dir='test/lnt',
-            alwaysUseLatest=True)
+        if testWithLNT:
+            f.addGetSourcecodeForProject(
+                project='lnt',
+                src_dir='test/lnt',
+                alwaysUseLatest=True)
+
         f.addGetSourcecodeForProject(
             project='test-suite',
             src_dir='test/test-suite',
@@ -409,7 +412,6 @@ def _getClangCMakeBuildFactory(
         cxx = 'clang-cl.exe'
         fc = 'flang-new.exe'
 
-
     ############# STAGE 2
     if useTwoStage:
         # We always cleanly build the stage 2. If the compiler has been
@@ -479,6 +481,10 @@ def _getClangCMakeBuildFactory(
     ############# TEST SUITE
     ## Test-Suite (stage 2 if built, stage 1 otherwise)
     if runTestSuite:
+        # Paths
+        sandbox = util.Interpolate('%(prop:builddir)s/test/sandbox')
+        test_suite_dir = util.Interpolate('%(prop:builddir)s/test/test-suite')
+
         compiler_path = stage1_install
         if useTwoStage:
             compiler_path=stage2_install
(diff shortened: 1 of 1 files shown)


omjavaid:
@gkistanova @DavidSpickett friendly ping!

DavidSpickett:
I seem to recall that a major reason to do this was that LNT would not run on the version of Python that first got a native Windows on Arm build. Now that https://github.com/llvm/llvm-lnt/commit/f987c1202562723621e4343e7d912304b058c7e3 has landed, is that a relevant concern?

On Linux we have been able to run the test suite with Python 3.10.12 since that change went in.

If there are other supporting reasons please list them.

(reproducing failures would be a bit easier certainly)

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -284,10 +285,12 @@ def _getClangCMakeBuildFactory(
...
     # so we wouldn't poison got_revision property.
     if runTestSuite or checkout_test_suite:
-        f.addGetSourcecodeForProject(
-            project='lnt',
-            src_dir='test/lnt',
-            alwaysUseLatest=True)
+        if testWithLNT:
+            f.addGetSourcecodeForProject(
+                project='lnt',
+                src_dir='test/lnt',
```
DavidSpickett, line 291: This is off topic really, but it has always annoyed me that repos get checked out to something other than the default name. It makes reproducing locally just a little bit more annoying.

There are others so I'll go through them at some point, no point changing this one now.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -494,103 +500,153 @@ def _getClangCMakeBuildFactory(
...
+
+            # Get latest built Clang (stage1 or stage2)
+            cc = util.Interpolate(f'%(prop:builddir)s/{compiler_path}/bin/{cc}')
+            cxx = util.Interpolate(f'%(prop:builddir)s/{compiler_path}/bin/{cxx}')
+
+            # LNT Command line (don't pass -jN. Users need to pass both --threads
+            # and --build-threads in nt_flags/test_suite_flags to get the same effect)
+            use_runtest_testsuite = len(nt_flags) == 0
+            if not use_runtest_testsuite:
+                test_suite_cmd = [python, lnt, 'runtest', 'nt',
```
DavidSpickett, line 517: I was looking at this earlier, wondering what nt even is. Nightly test?

The logic is a bit convoluted here but I see you're porting the existing style so it's ok for now.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -494,103 +500,153 @@ def _getClangCMakeBuildFactory(
...
-                               haltOnFailure=True,
-                               description=['running the test suite'],
-                               workdir='test/sandbox',
-                               logfiles={'configure.log'   : 'build/configure.log',
-                                         'build-tools.log' : 'build/build-tools.log',
-                                         'test.log'        : 'build/test.log',
-                                         'report.json'     : 'build/report.json'},
-                               env=test_suite_env))
+            f.addStep(WarningCountingShellCommand(name='build testsuite',
+                                                  command=ninja_cmd,
```
DavidSpickett, line 638: This will need to have "-k 0" added so we try to build everything possible (equivalent of https://github.com/llvm/llvm-lnt/pull/25).

But it's neater if we keep that for my PR which will be adding it to the LNT path.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -494,103 +500,153 @@ def _getClangCMakeBuildFactory(
...
+            f.addStep(LitTestCommand(
+                                name='test-suite',
+                                command=test_suite_cmd,
+                                haltOnFailure=True,
+                                description=['running the test suite'],
+                                workdir='test/sandbox',
+                                logfiles={'configure.log'   : 'build/configure.log',
+                                            'build-tools.log' : 'build/build-tools.log',
+                                            'test.log'        : 'build/test.log',
+                                            'report.json'     : 'build/report.json'},
```
DavidSpickett, line 596: What logs do we get from the cmake only test suite build? Is it just like a normal `ninja check-all`, stdio with the output, ending with a table of results?

zorg/buildbot/builders/ClangBuilder.py
DavidSpickett: No one in zorg ever passes nt_flags, but this PR wouldn't be the time to remove it anyway.

gkistanova:
This patch looks quite invasive. Did you consider using a different build factory for your purpose? Like [UnifiedTreeBuilder](https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/builders/UnifiedTreeBuilder.py) for example?

zorg/buildbot/builders/ClangBuilder.py
omjavaid: I agree with your point but changing directory name should be fairly silent.

zorg/buildbot/builders/ClangBuilder.py
omjavaid: LNT has a nt (nightly test) configuration. More on it can be found here https://llvm.org/docs/lnt/tests.html. I have not tested this LNT configuration myself.

zorg/buildbot/builders/ClangBuilder.py
omjavaid: As most buildbots run these test for validation I have kept the log to mimic logs output generated by llvm-project's ninja check-all

If we run ninja check it will actually run `llvm-lit -sv .` which generates logs like we get in llvm-project ninja check-all. 

For detailed logs We should run `llvm-lit .` with sv to get logs like following:
 
`compile_time: 0.0323
exec_time: 0.0000
hash: "a787a4de011d3c645d10ccb5efa9df57"
link_time: 0.0561
size: 25984
size..bss: 8
size..comment: 151
size..data: 24
size..data.rel.ro: 568
size..dynamic: 528
size..dynstr: 412
size..dynsym: 456
size..eh_frame: 1652
size..eh_frame_hdr: 404
size..fini: 20
size..fini_array: 8
size..gcc_except_table: 124
size..gnu.hash: 28
size..gnu.version: 38
size..gnu.version_r: 144
size..got: 48
size..got.plt: 120
size..init: 24
size..init_array: 8
size..interp: 27
size..note.ABI-tag: 32
size..plt: 224
size..rela.dyn: 1800
size..rela.plt: 288
size..rodata: 61
size..text: 5128
**********`

zorg/buildbot/builders/ClangBuilder.py
omjavaid: This is interesting. For now Clang builder does not have the option to pass -k option to ninja. I am still thinking on whether we should allow buildbots to configure what number tests should fail build before it should stop.
Otherwise we can just hardcode a value such as ninja -k 10 or 100 or more when we run LLVM testsuite.

omjavaid:
> This patch looks quite invasive. Did you consider using a different build factory for your purpose? Like [UnifiedTreeBuilder](https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/builders/UnifiedTreeBuilder.py) for example?

The diff looks a bit invasive but if we examine the code there are not many changes to existing LNT based builder logic. It is just a if condition separation between LNT and No LNT logic. The if condition indentation presents like a lot has changed.

Also I did consider other builders but ClangBuilder aligns well with our existing buildbots and infrastructure.

zorg/buildbot/builders/ClangBuilder.py
DavidSpickett: That part is fine, 99% of the time I'm looking at the "stdio" file anyway.

Is buildbot still able to parse the failures correctly and show the various failure types in the UI? (I forget if it parses the stdio or the json results file)

zorg/buildbot/builders/ClangBuilder.py
DavidSpickett: I think if you're anyone other than the committer of the change that broke the one test, `-k 0` is what you'd want. If I'm looking at a build as a maintainer I want to know if there are other failures I need to report.

That said, when we did have thousands of NOEXE, the UI only showed a limited number so `-k 10` might end up equivalent but I'd still say use `-k 0` because then we can look at the results in stdio or json to get the full picture.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -154,7 +154,7 @@ def getClangCMakeBuildFactory(
             testsuite_flags=None,
             submitURL=None,
             testerName=None,
-
+            testWithLNT=True,
```
DavidSpickett, line 157: `test` could mean `ninja check` or `llvm-test-suite`, perhaps change this to `testsuite_use_lnt` to match the other parameter? Then it's more clear what it changes.

zorg/buildbot/builders/ClangBuilder.py
DavidSpickett: And now I see you added it, cool.

DavidSpickett:
For me reading this it's not that it's that invasive but that it adds to an already gigantic function. While there are advantages to having it all in one place like that, this coding adding `if <100 lines> else <100 lines>` is not easy to read.

I would try breaking the if and else parts into standalone functions like `AddTestSuiteLNTSteps` and so on. Even if it means repeating a few lines between both.

DavidSpickett:
I just stumbled onto https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/builders/TestSuiteBuilder.py.

Perhaps that's what you were referring to Galina.

PR #246, opened 2024-07-29
Removing the ppc64be buildbots due to ppc64 linux OS EOL

kamaub:
RHEL 7.9, the only remaining officially supported Big Endian
Linux operating system, has reached its end of life. This patch
removes these buildbots due to their host and target arch
lack of OS security fixes and support.


amy-kwan (approved):
LGTM.

Nit on the description. Change "This patch removed these buildbots" to "This patch removes these buildbots".

kamaub:
Thanks you for the approval and nit @amy-kwan I have made the change. I do think it might be good to wait a bit for @gkistanova's input in case there is anything extra that needs to happen on the main server or anything related.

gkistanova (approved):
Thanks Kamau!
LGTM.

PR #247, opened 2024-07-30
[CUDA] Add libc and offload to the depends_on_projects

Artem-B:
Now that CUDA bots run libc-on-GPU tests, we want to trigger the tests on libc and offload changes, too.


jhuber6 (approved):

gkistanova (approved):
LGTM.

PR #249, opened 2024-08-11
Remove -x from shell scripts

vitalybuka:
It's too verbose.


goussepi:
Hi @vitalybuka , I find those verbose outputs useful though whenever I need to figure how to run my build and test as close as possible as the upstream bot, is it possible to keep the important invocations visible in the log? eg cmake and ninja invocations.
Many thanks!
Pierre

PR #252, opened 2024-08-22
Update LLVM WoA buildbots for better coverage

omjavaid:
This updates LLVM WoA bots with following changes:
1) Do not run stage1 testing on 2 stage clang bot. 
2) Remove redundant workers.
3) Remove -DCOMPILER_RT_BUILD_PROFILE=OFF
4) Add compiler-rt to single stage bot
5) Set clean=False for the 2stage bots
6) Set "-DCLANG_DEFAULT_LINKER=lld" for all WoA bots

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -320,10 +320,10 @@ all = [
                     clean=False,
                     checkout_flang=True,
                     checkout_lld=True,
-                    checkout_compiler_rt=False,
                     extra_cmake_args=[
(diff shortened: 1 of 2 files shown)


DavidSpickett:
Sorry I completely missed this.

> Set clean=False for the 2stage bots

Does this make a difference now that we use ccache? Also we may have to switch it back to clean if https://github.com/llvm/llvm-zorg/pull/266 works out.

You can leave it in here for now though.

DavidSpickett (approved):
LGTM

PR #253, opened 2024-08-22
Add WoA testsuite buildbot

omjavaid:
This is a follow up patch on #245 and #252
It adds a new builder called clang-arm64-windows-msvc-testsuite. It will
do stage1 and stage2 builds without running checks. Checks are already
covered by single stage and 2 stage bots.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -320,10 +320,10 @@ all = [
                     clean=False,
                     checkout_flang=True,
                     checkout_lld=True,
-                    checkout_compiler_rt=False,
                     extra_cmake_args=[
+                        "-DCLANG_DEFAULT_LINKER=lld",
                         "-DCMAKE_TRY_COMPILE_CONFIGURATION=Release",
-                        "-DLLVM_TARGETS_TO_BUILD='AArch64'",
+                        "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
                         "-DCMAKE_C_COMPILER_LAUNCHER=ccache",
                         "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache"])},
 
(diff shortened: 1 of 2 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -591,22 +591,42 @@
...
+    'builddir': "clang-arm64-windows-msvc-testsuite",
+    'factory' : ClangBuilder.getClangCMakeBuildFactory(
+                    vs="manual",
+                    checks=[],
+                    clean=False,
+                    checkout_flang=True,
+                    checkout_lld=True,
+                    runTestSuite=True,
+                    testWithLNT=False,
+                    testsuite_flags=["-DTEST_SUITE_SUBDIRS='Fortran'"],
```
DavidSpickett, line 606: Does enabling flang not just automatically do this?

buildbot/osuosl/master/config/builders.py
```diff
@@ -591,22 +591,42 @@
...
                     extra_cmake_args=[
+                        "-DCLANG_DEFAULT_LINKER=lld",
                         "-DCMAKE_TRY_COMPILE_CONFIGURATION=Release",
                         "-DCMAKE_C_COMPILER_LAUNCHER=ccache",
                         "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache",
-                        # FIXME: compiler-rt\lib\sanitizer_common\sanitizer_unwind_win.cpp assumes WIN64 is x86_64,
-                        #        so, before that's fixed, disable everything that triggers its build.
-                        "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
-                        "-DCOMPILER_RT_BUILD_PROFILE=OFF"])},
+                        "-DCOMPILER_RT_BUILD_SANITIZERS=OFF"])},
```
DavidSpickett, line 629: On Windows on Arm, we are only building compiler-rt for the compiler runtime part, correct? I see that all these are disabling the sanitizers.

buildbot/osuosl/master/config/workers.py
```diff
@@ -41,7 +41,6 @@ def get_all():
         create_worker("linaro-g3-04", max_builds=1),
 
         # AArch64 Windows Microsoft Surface X Pro
-        create_worker("linaro-armv8-windows-msvc-01", max_builds=1),
```
DavidSpickett, line 44: Be careful with the timing of when this change lands vs. when you remove the agent from the actual machine. I managed to get our server IP banned because I left some workers running that kept trying to connect to the build master.

If the machine is offline before this lands there should be no problem.

buildbot/osuosl/master/config/builders.py
```diff
@@ -591,22 +591,42 @@
                         "-DMLIR_RUN_ARM_SVE_TESTS=True",
                         "-DLLVM_LIT_ARGS='-v'"])},
 
+    {'name' : "clang-arm64-windows-msvc-testsuite",
+    'tags'  : ["clang"],
+    'workernames' : ["linaro-armv8-windows-msvc-03"],
+    'builddir': "clang-arm64-windows-msvc-testsuite",
+    'factory' : ClangBuilder.getClangCMakeBuildFactory(
+                    vs="manual",
+                    checks=[],
```
DavidSpickett, line 600: Where are you disabling the check stage 1 and 2 here? Also this doesn't seem to be a 2 stage, there's no useTwoStage.

PR #260, opened 2024-09-02
Make all clang-aarch64-full-2stage builds clean builds

DavidSpickett:
This is step 1 to addressing the rapidly expanding build queues we're seeing (#250).

The bot is currently on silent, and the plan is:
* Mark all builds as clean (we use ccache so it shouldn't be a big impact).
* Clear the existing requests.
* See what happens to the queue.

It's possible we will need a follow up to make collapsing work, but this is the first step either way.


antmox (approved):
LGTM

PR #263, opened 2024-09-11
[lldb][Arm] Use lld for linking and tests

DavidSpickett:
Since https://lab.llvm.org/buildbot/#/builders/18/builds/3578 the build has been failing with:
```
FAILED: lib/libclang-cpp.so.20.0git
: && /usr/local/bin/c++ -fPIC -fPIC -fno-semantic-interposition <...> -lm  /usr/lib/arm-linux-gnueabihf/libz.so && :
/usr/bin/ld.gold: internal error in open, at ../../gold/descriptors.cc:99
```

gold is no longer maintained and in my own experience hitting this same sort of error, my only solution was to switch to ld or lld.

There's no particular reason to use gold here I think it was just set that way before lld was useable on Arm.

Since we're linking clang with it we might as well use it in tests too, so make lld clang's default linker.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1311,6 +1311,7 @@ all = [
                         '-DLLVM_ENABLE_ASSERTIONS=True',
                         '-DLLVM_LIT_ARGS=-vj 4',
                         '-DLLVM_USE_LINKER=lld',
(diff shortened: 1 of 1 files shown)


omjavaid:
Should we also add -DCLANG_DEFAULT_LINKER=lld ? that way just built clang used during testsuite run will also use lld.

omjavaid (approved):

PR #266, opened 2024-09-24
Collapse clean and non clean build requests on clean builders

### LINKED ISSUE #250: Build requests piling up on workers

DavidSpickett:
Fixes #250 

Where build requests were piling up on slower builders because they were clean/don't clean/clean/ don't clean etc. and we could not merge those requests because of that difference.

Galina suggested that if we knew that the builder was going to clean the build first anyway, we could merge them. This commit implements that.

We do this by first setting a "clean" attribute on the builder's factory object, that we can find later via in the collapse callback.

So far only ClangBuilder passes this on but any builder can opt in by copying what it does.

In the collapse function when comparing properties we delete the "clean_obj" key from both sets of properties if we know that the builder's factory is always clean.

With some logging (not included in this commit) we can see it now collapses a clean/non-clean pair of requests:
```
collapseRequests
selfBuildSetProperties: {'scheduler': ('main:clang,clang-tools-extra,compiler-rt,flang,lld,llvm,mlir', 'Scheduler')}
otherBuildSetProperties: {'scheduler': ('main:clang,clang-tools-extra,compiler-rt,flang,lld,llvm,mlir', 'Scheduler'), 'clean_obj': (True, 'Change')}
Removing clean_obj from build set properties.
Build set properties were the same.
```

This has beeen tested by my colleague Omair Javaid on a test build master. Of course we only tested it with one of Linaro's builders, so it's possible we see other side effects in production.

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -278,7 +278,8 @@ def _getClangCMakeBuildFactory(
     f = LLVMBuildFactory(
             depends_on_projects=depends_on_projects,
             llvm_srcdir='llvm',
-            enable_runtimes=enable_runtimes)
+            enable_runtimes=enable_runtimes,
+            clean=clean)
 
     # Checkout the latest code for LNT
     # and the test-suite separately. Le's do this first,
--- a/zorg/buildbot/process/buildrequest.py
+++ b/zorg/buildbot/process/buildrequest.py
@@ -23,8 +23,8 @@ def collapseRequests(master, builder, req1, req2):
 
     # Fetch the buildset properties.
     selfBuildsetPoperties = yield \
-    master.db.buildsets.getBuildsetProperties(
-        str(req1['buildsetid'])
+        master.db.buildsets.getBuildsetProperties(
+            str(req1['buildsetid'])
         )
     otherBuildsetPoperties = yield \
         master.db.buildsets.getBuildsetProperties(
--- a/zorg/buildbot/process/buildrequest.py
+++ b/zorg/buildbot/process/buildrequest.py
@@ -22,14 +22,20 @@ def collapseRequests(master, builder, req1, req2):
         ('buildsets', str(req2['buildsetid'])))
 
     # Fetch the buildset properties.
-    selfBuildsetPoperties = yield \
-        master.db.buildsets.getBuildsetProperties(
-            str(req1['buildsetid'])
-        )
-    otherBuildsetPoperties = yield \
-        master.db.buildsets.getBuildsetProperties(
-            str(req2['buildsetid'])
-        )
+    selfBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
+        str(req1["buildsetid"])
+    )
+    otherBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
+        str(req2["buildsetid"])
+    )
+
+    # If the build is going to be a clean build anyway, we can collapse a clean
+    # build and a non-clean build.
+    if getattr(builder.config.factory, "clean"):
+        if 'clean_obj' in selfBuildsetPoperties:
+            del selfBuildsetPoperties["clean_obj"]
+        if 'clean_obj' in otherBuildsetPoperties:
+            del otherBuildsetPoperties["clean_obj"]
 
     # Check buildsets properties and do not collapse
     # if properties do not match. This includes the check
--- a/zorg/buildbot/process/buildrequest.py
+++ b/zorg/buildbot/process/buildrequest.py
@@ -31,7 +31,7 @@ def collapseRequests(master, builder, req1, req2):
 
     # If the build is going to be a clean build anyway, we can collapse a clean
     # build and a non-clean build.
-    if getattr(builder.config.factory, "clean"):
+    if getattr(builder.config.factory, "clean", False):
         if 'clean_obj' in selfBuildsetPoperties:
             del selfBuildsetPoperties["clean_obj"]
         if 'clean_obj' in otherBuildsetPoperties:
--- a/zorg/buildbot/process/buildrequest.py
+++ b/zorg/buildbot/process/buildrequest.py
@@ -22,12 +22,14 @@ def collapseRequests(master, builder, req1, req2):
         ('buildsets', str(req2['buildsetid'])))
 
     # Fetch the buildset properties.
-    selfBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
-        str(req1["buildsetid"])
-    )
-    otherBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
-        str(req2["buildsetid"])
-    )
+    selfBuildsetPoperties = yield \
+    master.db.buildsets.getBuildsetProperties(
+        str(req1['buildsetid'])
+        )
+    otherBuildsetPoperties = yield \
+        master.db.buildsets.getBuildsetProperties(
+            str(req2['buildsetid'])
+        )
 
     # If the build is going to be a clean build anyway, we can collapse a clean
     # build and a non-clean build.
--- a/zorg/buildbot/process/factory.py
+++ b/zorg/buildbot/process/factory.py
@@ -35,11 +35,15 @@ class LLVMBuildFactory(BuildFactory):
             & etc.
     """
 
-    def __init__(self, steps=None, depends_on_projects=None, hint=None, **kwargs):
+    def __init__(
+        self, steps=None, depends_on_projects=None, hint=None, clean=False, **kwargs
+    ):
         # Cannot use "super" here as BuildFactory is an old style class.
         BuildFactory.__init__(self, steps)
 
         self.hint = hint
+
+        self.clean = clean
         
         # Handle the dependencies.
         if depends_on_projects is None:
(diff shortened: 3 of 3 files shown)


joker-eph:
Why would we prevent from collapsing clean/non-clean builds?

DavidSpickett:
We already merge:
```
unclean
unclean
```
And
```
clean
clean
```

As the properties matched.

Now we want to merge:
```
clean
unclean
```
Assuming unclean merges into clean, and the result is also clean, we could do this for any builder even if it does not clean by default. In current production we wouldn't because the properties don't match.

```
unclean
clean
```
This we can only merge if the builder itself would clean anyway, or if we are able to edit the request after merging to add `clean_obj` Currently it's just a boolean collapse or not.

I will see if I can edit the collapsed request, then this can apply globally not just to builders marked clean.

zorg/buildbot/process/buildrequest.py
```diff
@@ -22,14 +22,20 @@ def collapseRequests(master, builder, req1, req2):
...
+        str(req2["buildsetid"])
+    )
+
+    # If the build is going to be a clean build anyway, we can collapse a clean
+    # build and a non-clean build.
+    if getattr(builder.config.factory, "clean"):
+        if 'clean_obj' in selfBuildsetPoperties:
+            del selfBuildsetPoperties["clean_obj"]
+        if 'clean_obj' in otherBuildsetPoperties:
+            del otherBuildsetPoperties["clean_obj"]
```
luporl, line 38: Does `master.db.buildsets.getBuildsetProperties` yields a buildset copy?

If not, wouldn't it be better to instead set `clean_obj` to `True` in both buildsets, just in case other part of the code relies on it?

omjavaid: Makes me wonder whether rest of the code depends on factory attribute "clean" or the build set properties clean_obj. From this change it looks like clean_obj was taking precedence over factory attribute hence the delete option was used.

Also I am thinking whether we should somehow restrict the number of builds that can be collapsed together. If queue is long may be only collapse n builds. Something that we can do in a later patch if anything drastic comes up during production builds.

omjavaid (approved):

DavidSpickett:
I'm going to see if it's feasible to modify the request that we collapse into, to address Medhi's question.

And to be clear, Galina has the deciding approval here.

zorg/buildbot/process/buildrequest.py
DavidSpickett: > Also I am thinking whether we should somehow restrict the number of builds that can be collapsed together. If queue is long may be only collapse n builds. Something that we can do in a later patch if anything drastic comes up during production builds.

I suspect this requires changes to Buildbot itself.

DavidSpickett:
> I'm going to see if it's feasible to modify the request that we collapse into, to address Mehdi's question.

This is the buildbot function that calls the callback I'm modifying in this PR: https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/buildrequest.py#L66

If we want to collapse unclean -< clean, we'd need to change this callback to return a new request with the `clean_obj` added. Or make the default some sort of superset of all options of both requests.

But I'd rather not get into changing Buildbot itself (or trying to monkeypatch it)

> Why would we prevent from collapsing clean/non-clean builds?

@joker-eph tell me if I've answered this or not. Might have gone on a sidetrack there.

DavidSpickett:
One sentence answer based on current buildbot behaviour:
"Collapsed requests retain the properties of the request being collapsed into, if that request is unclean we would lose the clean property of the collapsed request."

That is unless we knew that the builder would clean anyway, which is what this PR checks for.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -278,7 +278,8 @@ def _getClangCMakeBuildFactory(
     f = LLVMBuildFactory(
             depends_on_projects=depends_on_projects,
             llvm_srcdir='llvm',
-            enable_runtimes=enable_runtimes)
+            enable_runtimes=enable_runtimes,
+            clean=clean)
```
slydiman, line 282: What is the purpose of `clean` here? It seems LLVMBuildFactory (and base BuildFactory) does not use `clean`. I see only `TODO: Handle clean property and self.clean attribute.` in zorg/buildbot/process/factory.py, line 175.
I don't see any usage of `f.clean` in ClangBuilder.py too.

slydiman (approved):
LGTM, 
But it seems it is necessary to add similar `clean=clean` to LLDBBuilder.py, PollyBuilder.py, OpenMPBuilder.py and BOLTBuilder.py. Note most builders use UnifiedTreeBuilder which already contains `clean` property. See the full list of factories with `clean=True` in llvm-zorg/buildbot/osuosl/master/config/builders.py

zorg/buildbot/process/buildrequest.py
```diff
@@ -22,14 +22,20 @@ def collapseRequests(master, builder, req1, req2):
...
+    selfBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
+        str(req1["buildsetid"])
+    )
+    otherBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
+        str(req2["buildsetid"])
+    )
+
+    # If the build is going to be a clean build anyway, we can collapse a clean
+    # build and a non-clean build.
+    if getattr(builder.config.factory, "clean"):
```
slydiman, line 34: Change to `if getattr(builder.config.factory, "clean", False):`, then you can keep zorg/buildbot/process/factory.py unchanged.

slydiman (changes requested):
The default value in getattr() is necessary to avoid an exception if some factory does not have the `clean` property.

zorg/buildbot/process/buildrequest.py
DavidSpickett: I've fixed the default, I mistakenly thought it returned None like a dictionary .get would.

I don't see how that means I don't have to change factory.py though. Without that change no factory would have the clean attribute set on it.

Unless you mean I should split that change out so that this PR is purely adding the ability to collapse more, then a follow up enables it?

(which is a decent idea, it's still possible I'm breaking unrelated builders here)

slydiman: Unless `clang` is not specified as the parameter explicitly the following code in line 87 do the trick:
```
        # Preserve the rest of the given extra attributes if any, so we could
        # expand the factory later.
        for k,v in kwargs.items():
            setattr(self, k, v)
```

DavidSpickett: I see what you mean now, I didn't see that loop.

It looks like things we know we want to use are popped from kwargs and then that loop catches the remaining ones. So I've changed clean to pop from kwargs with a default False.

We could rely on that loop entirely but it's pretty cryptic. An explicit check implies it's something we care about and is a target for anyone searching for this later.

slydiman (approved):
Thanks

gkistanova (commented):
Thanks for working on this, David!

The patch looks good with a small nit pick. Please see my comments inline.
In the mean time I staged the patch to see it in action.

zorg/buildbot/process/buildrequest.py
```diff
@@ -22,14 +22,20 @@ def collapseRequests(master, builder, req1, req2):
...
     # Fetch the buildset properties.
-    selfBuildsetPoperties = yield \
-        master.db.buildsets.getBuildsetProperties(
-            str(req1['buildsetid'])
-        )
-    otherBuildsetPoperties = yield \
-        master.db.buildsets.getBuildsetProperties(
-            str(req2['buildsetid'])
-        )
+    selfBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
```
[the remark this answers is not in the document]
DavidSpickett: Done.

I didn't want to fix the formatting, just ran the usual lvm-project formatter on the new changes and it picked up too much.

zorg/buildbot/process/buildrequest.py
```diff
@@ -22,14 +22,20 @@ def collapseRequests(master, builder, req1, req2):
...
+        str(req1["buildsetid"])
+    )
+    otherBuildsetPoperties = yield master.db.buildsets.getBuildsetProperties(
```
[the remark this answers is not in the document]
DavidSpickett: Done.

DavidSpickett:
> In the mean time I staged the patch to see it in action.

I've cleared the queue for one of our affected builders, https://lab.llvm.org/staging/#/builders/75, and will see how the queue develops.

gkistanova (approved):
Thanks, David!
The patch works on the staging well.

vitalybuka:
> Is this on staging? 

Oh, I forgot to merge #257.

vitalybuka:
> Thanks, David! The patch works on the staging well.

It does not look like it works. https://lab.llvm.org/staging/#/builders/7
Is this rely on staging?

gkistanova:
> It does not look like it works. https://lab.llvm.org/staging/#/builders/7 Is this rely on staging?

You patch to make that builder clean was not on the staging yet. It is now.
I removed the current build queue, so we can see how it will go further on.

Let's wait for another day to make sure everything is fine. I'll be surprised if it is not, though.

gkistanova:
However, since `sanitizer-buildbot9` is offline, no build requests is going to be scheduled. Is there a way to get  that worker online on staging?

vitalybuka (approved):
https://lab.llvm.org/staging/#/builders/7 LGTM

gkistanova:
I think this is ready to merge.

It could be in production on Friday late night or Staurday morning.

PR #267, opened 2024-09-26
ClangBuilder: Use "flang" instead of "flang-new" for the test suite

DavidSpickett:
See https://discourse.llvm.org/t/proposal-rename-flang-new-to-flang/69462 for background.

This is to be landed once https://github.com/llvm/llvm-project/pull/110023 has landed.


luporl (approved):

gkistanova (approved):
LGTM.

PR #268, opened 2024-10-01
Remove OS version from build_deps

justice-adams-apple:
We have a fleet which has multiple OS versions, and we don't actually require that the OS == 14.1, so this check is really only causing false-positive errors to appear in the console logs


justice-adams-apple:
Example:
https://green.lab.llvm.org/job/llvm.org/job/llvm-coverage/63/

PR #270, opened 2024-10-08
[lldb] Flags for running Shell tests remotely on "lldb-remote-linux-ubuntu" builder

dzhidzhoev:
dotest.py flags are replaced with CMake flags introduced in https://github.com/llvm/llvm-project/pull/95986 to run Shell tests remotely.

Also, USE_LLVM_TOOLS is removed since it's implied by default that LLVM tools will be used for API tests since https://github.com/llvm/llvm-project/pull/109961.


vvereschaka (approved):
LGTM

PR #273, opened 2024-10-09
[CI] Using libgcc breaks Driver test

jplehr:
This makes using libunwind the default and work around a breakage in check-flang when setting libgcc as the default unwind lib.

(Should I remove the unrelated whitespace changes?)


Meinersbur:
What exactly is the error?

I think it is a courtesy to reviewers to remove whitespace noise (and also the git blame history).

jplehr:
Sure, I'll go ahead and update the patch to remove the whitespace noise.

The error can be found in the logs of this bot, together with the actual CMake invocation: https://lab.llvm.org/staging/#/builders/105

Meinersbur (approved):
So if I understand correctly, using libgcc for stack unwinding breaks the test that uses `-rtlib=compiler-rt` and expects that libgcc is not used anymore.

jplehr:
That is my understanding of the error, too.

PR #275, opened 2024-10-10
[Flang] Revert/Fix some bot changes.

jplehr:
For the time being, change the Flang bot config to something without libcxx, libunwind, compiler-rt.
This is an attempt to bring the bot back to green and guarantee some coverage.

The issues that showed up:
- Enabling libunwind as default unwind library leads to file-not-found-errors in the OpenMP tests.
- [edit] Adding libunwind also makes Offload believe the system is 32 bit and not build the offload library [/edit]
- Enabling compiler-rt and not libunwind leads to a failing build of some parts of archer.
- Removing libunwind and compiler-rt while enabling libcxx/libcxx-abi does lead to a CMake error that it wants to use libunwind, but that is not specified.


gkistanova (approved):
LGTM.

PR #277, opened 2024-10-12
Add clone of llvm-clang-x86_64-gcc-ubuntu builder that builds without assertions.

dyung:
This adds a builder which is a clone of our existing llvm-clang-x86_64-gcc-ubuntu builder, except without assertions enabled. This is useful for finding tests that are added/updated but rely on (usually debug) information that isn't available when assertions are not enabled. This should help to catch those issues.


dyung:
@gkistanova ping

gkistanova (approved):
LGTM.

PR #278, opened 2024-10-15
[zorg][lldb] Add new LLDB metrics bot

Michael137:
This patch adds a new job to collect LLDB metrics.

This is heavily based on the `debuginfo-statistics` job (but currently doesn't publish data to LNT).

Currently this job would do the following:
1. Check out the Clang 19.x release and build it
2. Use the LLDB and Clang from the `lldb-cmake-intel` job (not actually sure if that job publishes the right artifacts at the moment) to run the `run_lldb_metrics.sh` script.
3. Said script will attach LLDB to Clang/LLDB and run various commands. Then it will dump the `statistics dump` command to `stdout` (note we don't do any kind of averaging of these over multiple runs, since the metrics we care about **should** stable across runs). We also currently run these test-scenarios through `hyperfine` and dump the timing data. But maybe for a first attempt this isn't necessary.

### DIFF
--- a/zorg/jenkins/jobs/jobs/lldb-statistics
+++ b/zorg/jenkins/jobs/jobs/lldb-statistics
@@ -1,0 +1,166 @@
+#!/usr/bin/env groovy
+pipeline {
+    options {
+        disableConcurrentBuilds()
+
+        timeout(time: 12, unit: 'HOURS')
+    }
+
+    parameters {
+        string(name: 'LABEL', defaultValue: params.LABEL ?: 'macos-x86_64', description: 'Node label to run on')
+        string(name: 'GIT_SHA', defaultValue: params.GIT_REVISION ?: '*/release/19.x', description: 'Git commit to build.')
+        string(name: 'ARTIFACT', defaultValue: params.ARTIFACT ?: 'lldb-cmake-intel/latest', description: 'Clang/LLDB artifact to use')
+        booleanParam(name: 'CLEAN', defaultValue: params.CLEAN ?: false, description: 'Wipe the build directory?')
+    }
+
+    agent {
+        node {
+            label params.LABEL
+        }
+    }
+    stages {
+        stage('Checkout') {
+            steps {
+                script {
+                    if(params.CLEAN) {
+                        deleteDir()
+                    }
+                }
+
+                timeout(30) {
+                    dir('src/clang-19') {
+                        checkout([$class: 'GitSCM', branches: [
+                            [name: params.GIT_SHA]
+                        ], userRemoteConfigs: [
+                            [url: 'https://github.com/llvm/llvm-project.git']
+                        ], extensions: [
+                            [$class: 'CloneOption',
+                            noTags: true, timeout: 30]
+                        ]])
+                    }
+                    dir('llvm-zorg') {
+                        checkout([$class: 'GitSCM', branches: [
+                            [name: '*/main']
+                        ], userRemoteConfigs: [
+                            [url: 'https://github.com/llvm/llvm-zorg.git']
+                        ]])
+                    }
+                }
+            }
+        }
+        stage('Setup Venv') {
+            environment {
+               PATH="$PATH:/usr/bin:/usr/local/bin"
+            }
+            steps {
+                sh '''
+                   rm -rf venv
+                   python3 -m venv venv
+                   set +u
+                   source ./venv/bin/activate
+                   pip install -r ./llvm-zorg/zorg/jenkins/jobs/requirements.txt
+                   set -u
+               '''
+            }
+        }
+        stage('Fetch Artifact') {
+            environment {
+               PATH="$PATH:/usr/bin:/usr/local/bin"
+            }
+            steps {
+                withCredentials([string(credentialsId: 's3_resource_bucket', variable: 'S3_BUCKET')]) {
+                    sh """
+                        source ./venv/bin/activate
+                        echo "ARTIFACT=${params.ARTIFACT}"
+                        python llvm-zorg/zorg/jenkins/monorepo_build.py fetch
+                        ls $WORKSPACE/host-compiler/lib/clang/
+                        VERSION=`ls $WORKSPACE/host-compiler/lib/clang/`
+                    """
+                }
+            }
+        }
+        stage('Build') {
+            environment {
+               PATH="$PATH:/usr/bin:/usr/local/bin"
+               SRC="$WORKSPACE/src"
+               BUILD="$WORKSPACE/clang-19-build"
+               TEST="$WORKSPACE/test"
+               RESULTS="$WORKSPACE/results"
+               CC="$WORKSPACE/host-compiler/bin/clang"
+               CXX="$WORKSPACE/host-compiler/bin/clang++"
+               HISTORIC_COMPILER="clang-19"
+            }
+            steps {
+                withCredentials([string(credentialsId: 's3_resource_bucket', variable: 'S3_BUCKET')]) {
+                    sh '''
+                       source ./venv/bin/activate
+
+                       cd src/clang-19
+                       git tag -a -m "First Commit" first_commit 97724f18c79c7cc81ced24239eb5e883bf1398ef || true
+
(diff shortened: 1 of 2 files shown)


Michael137:
First stab at this. The plan is to collect metrics that relate to type completion. E.g.,:
1. How many types did we fully resolve?
2. How many types did we keep as forward declarations?
3. How many definitions did ASTImporter import?
4. How many FindTypes/FindNamespace/FindFunctions calls did we perform?
5. How many object files did we scan to find a type?

These don't all exist yet, but the idea is to add them to the `statistics dump` command.

zorg/jenkins/jobs/util/run_lldb_metrics.sh
```diff
@@ -0,0 +1,180 @@
...
+    "${LLVM_BUILD_DIR}/bin/lldb -o 'b ClangASTSource::FindExternalVisibleDeclsByName' -o run -o 'expr clang_decl_name.getAsString()'"
+
+profile_lldb \
+    "expr_re_eval" \
+    "${LLVM_BUILD_DIR}/bin/lldb -o 'b ClangASTSource::FindExternalVisibleDeclsByName' -o run -o 'expr m_active_lexical_decls' -o 'expr m_active_lexical_decls'"
+
+profile_lldb \
+    "expr_two_stops" \
+    "${LLVM_BUILD_DIR}/bin/lldb -o 'b ClangASTSource::FindExternalVisibleDeclsByName' -o run -o 'expr m_active_lexical_decls' -o 'b CodeGeneratorImpl::HandleTranslationUnit' -o 'br del 1' -o c -o 'expr Builder' -o 'expr Ctx.getLangOpts()'"
+
```
adrian-prantl, line 173: Could you add some comments explaining for future maintainers why these various breakpoints and expressions were chosen and what is being tested?

For example, I expect the first one to trigger all sorts of symbol file loading, the second one to test the same expression with a warm cache, one finding a templated type with lots of false positives that appears 1000 times, ...

adrian-prantl (commented):
This is going to be incredibly useful, thanks!

zorg/jenkins/jobs/util/run_lldb_metrics.sh
Michael137: Took a stab at it. Wasn't sure of the level of detail. Let me know if i should elaborate more here

Michael137:
> * (where) will we be able to see the results of these benchmarks?

Currently I just dump them to the console. The idea in the near future is to publish the data to something like LNT (though that currently seems to be down), and plot some sort of time series out of it.

> * When benchmarking a debugger, there are many moving parts: a) the debugger itself; b) the code being debugged (inferior); c) the compiler compiling the inferior (; and possibly d) compiler compiling the debugger). Moving all four makes it hard to interpret the results. Based on the mentions of "historic compilers" in the patch, I'm deducing that you're trying to fix some of these, but I wasn't able to figure out which ones. Can you tell me which of these are fixed?

Good point, I'll try to clarify this in the pipeline definition.

The debugger/compiler that we're debugging (in the `HISTORIC_BUILD_DIR`) is pinned to the llvm-19.x release (which seemed like a good starting point for something stable).

The "host" compiler/LLDB is taken from whatever the LLDB incremental built/used (I haven't checked that those artifacts are available, but would be nice if we could re-use that). We use that compiler to build the "historic" Clang/LLDB. The metrics we collect are from that "host" LLDB that we fetched.

Do we want to pin the version of the compiler that compiles the historic LLDB/Clang? I guess if something changed in the way that we built the accelerator tables/etc. we would want that to reflect in the statistics.

labath:
> > * (where) will we be able to see the results of these benchmarks?
> 
> Currently I just dump them to the console. The idea in the near future is to publish the data to something like LNT (though that currently seems to be down), and plot some sort of time series out of it.

Got it. Thanks.

> 
> > * When benchmarking a debugger, there are many moving parts: a) the debugger itself; b) the code being debugged (inferior); c) the compiler compiling the inferior (; and possibly d) compiler compiling the debugger). Moving all four makes it hard to interpret the results. Based on the mentions of "historic compilers" in the patch, I'm deducing that you're trying to fix some of these, but I wasn't able to figure out which ones. Can you tell me which of these are fixed?
> 
> Good point, I'll try to clarify this in the pipeline definition.
> 
> (a) The "host" compiler/LLDB is taken from whatever the LLDB incremental built/used (I haven't checked that those artifacts are available, but would be nice if we could re-use that). The metrics we collect are from that "host" LLDB that we fetched.

One argument for *not* fetching those is that you might want to use different build options for each. E.g. the incremental build bot might want to enable assertions or stuff, whereas the benchmarking bot might not.
> 
> (b) The debugger/compiler that we're debugging (in the `HISTORIC_BUILD_DIR`) is pinned to the llvm-19.x release (which seemed like a good starting point for something stable)
> 
> (c) We use the compiler from (a) to build the "historic" Clang/LLDB.
> 
> (d) The compiler compiling the debugger in (a) is the clang produced by the clang-stage2 buildbot

:+1: 


> 
> We could alternatively choose not to re-use the artifacts from other buildbots and instead build a brand new LLDB/Clang from top-of-tree using a pinned version of Clang. In that case (b) and (d) would be stable, while (a) and (c) followed top-of-tree. That does seem like a more maintainable situation (at the cost of rebuilding Clang/LLDB more often)

I think both of these are reasonable choices, and its up to you to choose which one makes most sense for your use case. I'm interesting in the details just so that I know how to interpret the results.

zorg/jenkins/jobs/jobs/lldb-statistics
```diff
@@ -0,0 +1,187 @@
...
+    options {
+        disableConcurrentBuilds()
+
+        timeout(time: 12, unit: 'HOURS')
+    }
+
+    parameters {
+        string(name: 'LABEL', defaultValue: params.LABEL ?: 'macos-x86_64', description: 'Node label to run on')
+        string(name: 'GIT_SHA', defaultValue: params.GIT_REVISION ?: '*/release/19.x', description: 'Git commit to build.')
+        string(name: 'ARTIFACT', defaultValue: params.ARTIFACT ?: 'lldb-cmake-intel/latest', description: 'Clang/LLDB artifact to use')
```
justice-adams-apple, line 31: I don't think this artifact exists right now

Michael137: Ah that was my worry. What would be the process of making it available? The only reference to publishing the artefacts that I could find is:
```
echo "ARTIFACT=$JOB_NAME/clang-d$GIT_DISTANCE-g$GIT_SHA-t$BUILD_ID-b$BUILD_NUMBER.tar.gz" >> build.properties```
in `jobs/clang-stage1-RA`. Is it some combination of a Jenkins configuration and this `build.properties`?

File: zorg/jenkins/jobs/jobs/lldb-statistics [hunk 3]
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env groovy
+
```
adrian-prantl, line 2: Can you add a top-level comment along the lines of:

This pipeline collects LLDB performance data by by running LLDB benchmarks while debugging a stable version of clang-19. ...

zorg/jenkins/jobs/jobs/lldb-statistics
```diff
@@ -0,0 +1,187 @@
...
+// These are the LLVM build directories involved:
```
adrian-prantl, line 3: Which reminds me: You should make sure to compile clang-19 with itself (basically the stage2 cmake cache in clang/cmake/caches) if that's not already happening since we want the binary to be as stable as possible.

adrian-prantl (approved):
I think this is great as a first iteration!

zorg/jenkins/jobs/jobs/lldb-statistics
Michael137: Good point. Updated to a 2-stage build. I'm using the `CLANG_ENABLE_BOOTSTRAP` variable instead of the `Apple-stage2.cmake`, because iiuc the latter would build an optimized build.

PR #280, opened 2024-10-16
Fix toolchain's sysroot for lldb-remote-linux-ubuntu builder.

vvereschaka:
Pass a worker specific property value as a toolchain's sysroot path for the toolchain build stage 1 of the builder.


gkistanova (approved):
LGTM.

PR #282, opened 2024-10-17
[GHA] Add GitHub Action to run the tests for llvm-zorg itself

asb:
This adds a GitHub action to at least run the very minimal lit tests we have for zorg itself. With appropriate repository settings (perhaps it just works as-is? I've tested a variant on a downstream fork), it should provide pass/fail for both PRs and commits.

The GH Actions in the main llvm-repos tend to be gated on `if: github.repository_owner == 'llvm'`, which I've copied across here.

There's a question mark about how to handle `test/jenkins/test_build.py`. I've marked it XFAIL as a command-line `lit` option, as the test ends up trying to call `xcrun` which has no hope of succeeding on Linux. But perhaps the test is fixable, or it would be better to mark the test itself XFAIL. Either way, I think the GH Action in the PR is definitely an improvement over the status quo.

### DIFF
--- a/.github/workflows/lit-tests.yml
+++ b/.github/workflows/lit-tests.yml
@@ -1,4 +1,4 @@
-name: Run llvm-zorg's lit tests
+name: llvm-zorg testing
 
 on: [push, pull_request]
 
@@ -6,7 +6,8 @@ permissions:
   contents: read
 
 jobs:
-  test:
+  check_zorg:
+    name: llvm-zorg lit tests
     if: github.repository_owner == 'llvm'
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/lit-tests.yml
+++ b/.github/workflows/lit-tests.yml
@@ -1,0 +1,21 @@
+name: Run llvm-zorg's lit tests
+
+on: [push, pull_request]
+
+permissions:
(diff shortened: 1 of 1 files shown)


asb:
Here's an example of the check working on a fork of this repo https://github.com/asb/llvm-zorg/actions/runs/11383050481/job/31667868020

.github/workflows/lit-tests.yml
```diff
@@ -0,0 +1,21 @@
+name: Run llvm-zorg's lit tests
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  test:
```
vvereschaka, line 9: may be it is optimal to change this line `test:` -> `Zorg-Lit-tests`. This is going to be a title of the test in the `checks` section as far as I understood.

vvereschaka: We could add the additional jobs later, something like `Buildbot-checkconfig`, `Buildbot-pylint' or `Zorg-pylint` & etc.

asb: I think the top-level name is the one that's most visible in the UI, but good point we could be more explicit. I've tweaked so it all shows like below:

image

PR #284, opened 2024-10-22
Run cmake install to build package

justice-adams-apple:
We need to install so we can actually create the binary dir which is then converted to the tarball


Michael137 (approved):
LGTM

PR #285, opened 2024-10-23
Set permissions on both install and build binary before uploading

justice-adams-apple:
call LLDB workaround correctly on binary in the install directory as well as the build directory


justice-adams-apple:
cc @Michael137

Michael137 (approved):
Thanks for fixing! Lets give it a shot

PR #289, opened 2024-10-30
Add local testing mode to the buildmaster config

asb:
This makes it (relatively) easy to test a builder setup locally. The buildmaster and the web interface should be bound only to local interfaces for security reasons (you can use ssh port forwarding if wanting to run on a server).

---
This patch ended up fairly small, but it needed quite a lot of fiddling about to get something that works and requires such minimal changes. I assume everyone has their own flows or downstream patches for testing with a local buildmaster, but I want to make it substantially easier for anyone to do this without repeating that effort. With this change, you just need to set the BUILDMASTER_TEST environment variable when starting the buildmaster, and it will bind to local interfaces and default to a "test" password for workers. The web interface is configured with no auth needed in this case.

I believe that any notification generation is disabled by default as `getReporters()` in `buildbot/osuosl/master/config/status.py` returns an empty array unless a config option is_production is explicitly set. It would be helpful if people more familiar with the codebase could also confirm there's no chance this will try to send unwelcome emails.

Open issue:
* The LLVM buildmaster seems to be running off a downstream buildbot fork that isn't publicly available. The issue I have running with buildbot 3.11.7 is the length of some step names overflow the limit (see [upstream issue](https://github.com/buildbot/buildbot/issues/3414)). I hack around this locally by removing the call to check_param_length in `site-packages/buildbot/process/buildstep.py` in my venv (running with an sqlite database, I don't believe the width is enforced anyway).

I intend to pair this with a section on local testing of a build config in <https://llvm.org/docs/HowToAddABuilder.html>, but the basic gist of using this is:
* python -m venv myenv
* source myenv/bin/activate
* pip install buildbot{,-console-view,-grid-view,-waterfall-view,-worker,-www}==3.11.7 urllib3
    * Ideally we'd use requirements.txt, but the inacessible buildbot fork is a barrier
* Hack site-packages/buildbot/process/buildstep.py to remove check for step name length (obviously it would be good to remove this requirement...)
* cd to buildbot/osuosl/master
* BUILDMASTER_TEST=1 buildbot checkconfig
* BUILDMASTER_TEST=1 buildbot upgrade-master . # needed to init the db first time
* BUILDMASTER_TEST=1 buildbot start --nodaemon .
* Confirm you see the web UI at http://localhost:8011
* Then create a buildbot worker (or workers) with the same name as one of the configured workers, password 'test', and set to connect to localhost:9990
* You can either wait until the poller sets off a build, or force one in the web UI.

### DIFF
--- a/buildbot/osuosl/master/master.cfg
+++ b/buildbot/osuosl/master/master.cfg
@@ -26,11 +26,18 @@ c['buildbotNetUsageData'] = None
 import config
 reload(config)
 
+# Detect if the BUILDMASTER_TEST environment variable is set and non-zero.
+# This will be used to enable a local testing mode.
+buildmaster_test = os.environ.get('BUILDMASTER_TEST')
+test_mode = bool(buildmaster_test) and buildmaster_test != '0'
+config.options['Internal'] = {}
+config.options['Internal']['test_mode'] = str(test_mode)
+
 ####### Workers
 
 c['workers'] = config.workers.get_all()
 
-c['protocols'] = {'pb': {'port': 9990}}
+c['protocols'] = {'pb': {'port': "tcp:9990:interface=127.0.0.1" if test_mode else 9990}}
 
 ####### CHANGESOURCES
 
@@ -94,25 +101,29 @@ c['services'] = config.status.getReporters()
 
 ####### PROJECT IDENTITY
 
-c['title'] = config.options.get('Buildbot', 'title')
-c['titleURL'] = config.options.get('Buildbot', 'title_url')
-c['buildbotURL'] = config.options.get('Buildbot', 'my_url')
+c['title'] = "LLVM Buildbot (local testing mode)" if test_mode else config.options.get('Buildbot', 'title')
+c['titleURL'] = "http://localhost:8011" if test_mode else config.options.get('Buildbot', 'title_url')
+c['buildbotURL'] = "http://localhost:8011" if test_mode else config.options.get('Buildbot', 'my_url')
 
 # minimalistic config to activate new web UI
-c['www'] = dict(port=8011,
-                plugins=dict(waterfall_view={}, console_view={}, grid_view={}), # TODO: badges
-                default_page='console',
-                auth=config.auth.getAuth(),
-                authz=config.auth.getAuthz(),
-                #logRotateLength=
-                #maxRotatedFiles=
-                #versions=
+www_config = dict(port="tcp:8011:interface=127.0.0.1" if test_mode else 8011,
+                  plugins=dict(waterfall_view={}, console_view={}, grid_view={}), # TODO: badges
+                  default_page='console',
+                  #logRotateLength=
+                  #maxRotatedFiles=
+                  #versions=
             )
+
+if not test_mode:
+    www_config['auth'] = config.auth.getAuth()
+    www_config['authz'] = config.auth.getAuthz()
+
+c['www'] = www_config
 
 ####### DB URL
 
 c['db'] = {
-    'db_url' : config.options.get('Database', 'db_url'),
+    'db_url' : "sqlite:///state.sqlite" if test_mode else config.options.get('Database', 'db_url'),
 }
 
 ####### RESOURCE USAGE
--- a/buildbot/osuosl/master/master.cfg
+++ b/buildbot/osuosl/master/master.cfg
@@ -101,7 +101,7 @@ c['services'] = config.status.getReporters()
 
 ####### PROJECT IDENTITY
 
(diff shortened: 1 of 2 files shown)


DavidSpickett:
I know zero about doing this because I always ask @omjavaid to do it for me :rofl: 

It would be wonderful to have this working. The long cycle times working with buildbots can be difficult to deal with so more local testing is always welcome.

buildbot/osuosl/master/master.cfg
```diff
@@ -26,11 +26,18 @@ c['buildbotNetUsageData'] = None
...
+test_mode = bool(buildmaster_test) and buildmaster_test != '0'
+config.options['Internal'] = {}
+config.options['Internal']['test_mode'] = str(test_mode)
+
 ####### Workers
 
 c['workers'] = config.workers.get_all()
 
-c['protocols'] = {'pb': {'port': 9990}}
+c['protocols'] = {'pb': {'port': "tcp:9990:interface=127.0.0.1" if test_mode else 9990}}
```
omjavaid, line 40: I not 100% sure on this but I believe putting interface=127.0.0.1 will make test buildmaster inaccessible from the outside world which is not always the case. For example, if we are running this test master in a cloud vps and want our out of network workers to have access to it.

buildbot/osuosl/master/master.cfg
```diff
@@ -94,25 +101,29 @@ c['services'] = config.status.getReporters()
...
-                #versions=
+www_config = dict(port="tcp:8011:interface=127.0.0.1" if test_mode else 8011,
+                  plugins=dict(waterfall_view={}, console_view={}, grid_view={}), # TODO: badges
+                  default_page='console',
+                  #logRotateLength=
+                  #maxRotatedFiles=
+                  #versions=
             )
 
+if not test_mode:
```
omjavaid, line 117: If we dont get auth from github, does it mean that all worker are accessible to us or are there other changes required for allowing test buildmaster owner to be the owner of all the connected workers.

omjavaid (commented):
This is a very good idea. I always had to keep these changes synced up.
Just some inline comments.

buildbot/osuosl/master/master.cfg
asb: That's correct, my thinking is to default to something that's as secure as possible. I use ssh port forwarding in this case, but I think `socat` could also work if you just want to quickly expose the port to the outside world with no auth needed. Or of course editing the command line yourself.

I can document ssh port forwarding and/or socat for this and the web interface in the followon patch to https://llvm.org/docs/HowToAddABuilder.html that describes testing mode.

Perhaps I'm being over-cautious? But on the other hand:
* This mode uses a default password for connections, meaning it's easy to connect to the buildmaster and thus exploitation is much more plausible vs a production buildmaster where attackers shouldn't be able to connect
* People often run a mode like this on developer machines or within secure networks, so the potential impact is high vs a production buildmaster where (hopefully!) time and thought have gone into appropriate permissions / sandboxing / firewalling.
* For most cases where you're interactively testing a buildmaster and set of builders, I'd imagine using ssh port forwarding does the job fine.

Anyway, that's my thinking - other suggestions welcome!

buildbot/osuosl/master/master.cfg
asb: Leaving out both auth and authz fields in the config dict seems to result in being able to access the web interface with no auth needed, and anyone accessing the interface being able to force builds and clear queues. Leaving out 'auth' but leaving in 'authz' means you can't do those commands.

Are there any other actions I might have missed I should test?

The `buildbotURL` also needs to be correct or else you get "invalid origin" errors when clicking those buttons in the web UI. This is another case ssh port forwarding helps if you happened to be testing on a server - as you can have it so `localhost:8011` forwards to 8011 on the loopback interface on the remote machine.

asb:
Firstly: I've submitted a fix for the issue of some configs having overly long step names (causing an error upon startup) to https://github.com/llvm/llvm-zorg/pull/293, and chosen to just apply this fix universally rather than trying to gate it on "testing mode" or not.

Secondly: Do let me know if there is any feedback people feel is unaddressed? I'd love to move forwards with this and an associated docs update.

asb:
Matching llvm documentation PR is now up here https://github.com/llvm/llvm-project/pull/115024 which includes instructions on how ssh port forwarding can be used (which I think provides further clarity beyond what I suggested before in response to @omjavaid's questions - thanks for taking a look!).

omjavaid (approved):
This looks good to me. I have tested it locally and it works out of the box which is great.
Please consider merging it along with your changes to  https://llvm.org/docs/HowToAddABuilder.html 
Thanks Alex!

PR #291, opened 2024-10-31
[clang-3.4-debuginfo-statistics] Fix the LNT submissions run order

azharudd:
The run order should be the GIT_DISTANCE of the stage 1 compiler which is being used to build the historic compiler.

### DIFF
--- a/zorg/jenkins/jobs/jobs/clang-3.4-debuginfo-statistics
+++ b/zorg/jenkins/jobs/jobs/clang-3.4-debuginfo-statistics
@@ -11,7 +11,7 @@ pipeline {
 
         string(name: 'GIT_SHA', defaultValue: params.GIT_REVISION ?: '*/release/13.x', description: 'Git commit to build.')
 
-        string(name: 'ARTIFACT', defaultValue: params.ARTIFACT ?: 'clang-stage1-RA/latest', description: 'Clang artifact to use')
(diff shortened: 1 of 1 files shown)


zorg/jenkins/jobs/jobs/clang-3.4-debuginfo-statistics
```diff
@@ -76,7 +89,7 @@ pipeline {
                 withCredentials([string(credentialsId: 's3_resource_bucket', variable: 'S3_BUCKET')]) {
                     sh """
                         source ./venv/bin/activate
-                        echo "ARTIFACT=${params.ARTIFACT}"
+                        export $(grep ARTIFACT last_good_build.properties)
```
justice-adams-apple, line 92: I this still downloading the toolchain package?

justice-adams-apple (approved):
LGTM pending small question

Michael137:
Thanks for addressing this! Makes sense to me. I found this bit of the pipeline suspicious when I was looking into doing the same for the lldb-statistics bot.

I'll defer to @justice-adams-apple for the review

zorg/jenkins/jobs/jobs/clang-3.4-debuginfo-statistics
azharudd: Yes, it should. It points the `ARTIFACT` to the one listed in `last_good_build.properties` and that is fetched in the next line.

PR #297, opened 2024-11-07
Add a new LLDB remote-linux builder on Windows host (lldb-remote-linux-win).

vvereschaka:
Also add a new worker for the builder - as-builder-10.


gkistanova (approved):
Thanks, Vlad!
LGTM.

PR #298, opened 2024-11-07
Fix a few shellcheck warnings in sanitizer buildbot script

arichardson:


asb (approved):
All standard stuff, LGTM. The output of `du -hs` will change slightly (it will include the ./ prefix) but it seems it's not parsed directly in this script anyway.

PR #301, opened 2024-11-10
Rename BUILDMASTER_TEST to BUILDBOT_TEST

asb:
BUILDBOT_TEST is sufficiently descriptive (after all, 'buildbot' is the package name and the name of the binary invoked to create a buildmaster) and it's possible upstream may later remove the buildmaster terminology <https://github.com/buildbot/buildbot/issues/5382> or alternatively we might move away from it downstream. As such, it seems prudent to use a name that would be unaffected.

----

This came partially from discussion at https://github.com/llvm/llvm-project/pull/115024 - I figure tweaking the name is a lot easier now, before it's fully documented/advertised and part of people's workflows, than later.

### DIFF
--- a/buildbot/osuosl/master/master.cfg
+++ b/buildbot/osuosl/master/master.cfg
@@ -26,10 +26,10 @@ c['buildbotNetUsageData'] = None
 import config
 reload(config)
 
-# Detect if the BUILDMASTER_TEST environment variable is set and non-zero.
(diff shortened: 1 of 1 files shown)


asb:
> LGTM but if this variable existed before your changes get the ok from Galina too.

No, I added it in #289 and there shouldn't be any users as it's only just landed and the docs didn't land yet.

PR #304, opened 2024-11-11
[AMDGPU][Flang] Adds MLIR as dependency

jplehr:
Since flang depends on MLIR, depend on MLIR for changes to catch problems early.


mjklemm (commented):
LGTM, nit: maybe put the dependency in order (so, `mlir` goes before `flang`)?

mjklemm (approved):
LGTM

PR #306, opened 2024-11-12
Make Linaro 2 stage builders use clean builds

DavidSpickett:
To take advantage of https://github.com/llvm/llvm-zorg/pull/266. We use ccache on all our bots so this will not add much overhead.

The SVE bots are fast enough that the buildup of build requests is not really a problem, but for consistency I've marked them clean so all our 2 stage builds are clean.


omjavaid (approved):
Looks fine to me

omjavaid:
> @omjavaid in particular the Windows 2 stage bot. IIRC it does use ccache so it should be fine?
+1

PR #311, opened 2024-11-19
Add Linaro Graviton 4 workers

DavidSpickett:
Aside from the new CPU and different core/RAM amounts, these will be the same as our existing Gravition 3 ("g3") workers.

Builders to use these coming in a separate PR.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -473,7 +473,8 @@ all = [
     # problems sooner rather than later.
     {'name' : "clang-aarch64-sve-vla",
     'tags'  : ["clang"],
-    'workernames' : ["linaro-g3-01", "linaro-g3-02", "linaro-g3-03", "linaro-g3-04"],
+    'workernames' : ["linaro-g3-01", "linaro-g3-02", "linaro-g3-03", "linaro-g3-04",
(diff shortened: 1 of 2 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -473,7 +473,8 @@
     # problems sooner rather than later.
     {'name' : "clang-aarch64-sve-vla",
     'tags'  : ["clang"],
-    'workernames' : ["linaro-g3-01", "linaro-g3-02", "linaro-g3-03", "linaro-g3-04"],
+    'workernames' : ["linaro-g3-01", "linaro-g3-02", "linaro-g3-03", "linaro-g3-04",
+                     "linaro-g4-01", "linaro-g4-02"],
```
maxim-kuvyrkov, line 477: I think, it's dangerous to run the same builder on workers with different hardware.  I suggest we add clang-aarch64-sve2-vla and clang-aarch64-sve2-vla-2stage builders.
The intention of running buildbots both on Graviton3 and Graviton4 machines is to catch bugs when SVE vectorization works on one piece of hardware, but fails on another.

DavidSpickett: Ok I will reduce this PR to the workers only and do the builders separately.

gkistanova (approved):
LGTM.

PR #317, opened 2024-11-20
Add AArch64 SVE2 Vector Length Agnostic builders

DavidSpickett:
These generate code that uses the SVE2 extension, as opposed to our existing builders that only use SVE. Therefore they will run on our Graviton 4 workers only, which are Neoverse v2 cores with SVE and SVE2.

The optimisation guide:
https://github.com/aws/aws-graviton-getting-started/blob/main/c-c++.md

Suggests -mcpu=neoverse-v2 for best performance, but here I'm using it just so we get the SVE2 feature enabled. If we used the balanced -mcpu=neoverse-512tvb, we wouldn't get SVE2 without fiddly extra flags.

```
$ ~/clang+llvm-18.1.8-aarch64-linux-gnu/bin/clang /tmp/test.c -o /dev/null -mcpu=neoverse-v2 -###
<...> "-target-feature" "+sve" "-target-feature" "+sve2-bitperm" "-target-feature" "+sve2" <...>
```

I have not added vector length settings as these builders will generate scalable ("length agnostic" - LA) code.

Otherwise, the builders have identical setups to the existing SVE VLA builders.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -464,6 +464,9 @@ all = [
                         "-DMLIR_RUN_ARM_SME_TESTS=True",
                         "-DARM_EMULATOR_EXECUTABLE=qemu-aarch64"])},
 
+    # All SVE (as opposed to SVE2) builders are using optimisation flags
+    # for Graviton 3 "balanced" from
+    # https://github.com/aws/aws-graviton-getting-started/blob/main/c-c++.md.
 
     # AArch64 Clang+LLVM+RT+LLD check-all + flang + test-suite +
     # mlir-integration-tests w/SVE-Vector-Length-Agnostic Note that in this and
@@ -566,7 +569,10 @@ all = [
                         "-DMLIR_RUN_ARM_SVE_TESTS=True",
                         "-DLLVM_LIT_ARGS='-v'"])},
 
-    # Note that this is SVE2 as opposed to SVE.
+    # All SVE2 builders are using optimisation flags for Graviton 4 "performance" from
+    # https://github.com/aws/aws-graviton-getting-started/blob/main/c-c++.md
+    # (using balanced would not enable the SVE2 extension).
+
     {'name' : "clang-aarch64-sve2-vla",
     'tags'  : ["clang"],
     'workernames' : ["linaro-g4-01", "linaro-g4-02"],
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -566,6 +566,53 @@ all = [
                         "-DMLIR_RUN_ARM_SVE_TESTS=True",
                         "-DLLVM_LIT_ARGS='-v'"])},
 
+    # Note that this is SVE2 as opposed to SVE.
+    {'name' : "clang-aarch64-sve2-vla",
+    'tags'  : ["clang"],
+    'workernames' : ["linaro-g4-01", "linaro-g4-02"],
+    'builddir': "clang-aarch64-sve2-vla",
+    'factory' : ClangBuilder.getClangCMakeBuildFactory(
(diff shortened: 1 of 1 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -566,6 +566,55 @@
...
+                    testsuite_flags=[
+                        '--cppflags', '-mcpu=neoverse-v2 -mllvm -scalable-vectorization=preferred -mllvm -treat-scalable-fixed-error-as-warning=false -O3',
+                        '--threads=32', '--build-threads=32'],
+                    extra_cmake_args=[
+                        "-DCMAKE_C_FLAGS='-mcpu=neoverse-v2'",
+                        "-DCMAKE_CXX_FLAGS='-mcpu=neoverse-v2'",
+                        "-DLLVM_ENABLE_LLD=True",
+                        "-DMLIR_INCLUDE_INTEGRATION_TESTS=True",
+                        "-DMLIR_RUN_ARM_SVE_TESTS=True",
+                        "-DLLVM_LIT_ARGS='-v'"])},
```
asb, line 590: ClangBuilder.py has logic to set `-DLLVM_LIT_ARGS` itself and defaults to `-v`. Is there a reason to override it?

DavidSpickett:
I've also confirmed that at least right now, these configurations are green. They will be silent to begin with anyway, so that we can work out any capacity issues.

gkistanova (approved):
LGTM.

buildbot/osuosl/master/config/builders.py
```diff
@@ -566,6 +566,55 @@
...
+    'factory' : ClangBuilder.getClangCMakeBuildFactory(
+                    clean=False,
+                    checkout_flang=True,
+                    runTestSuite=True,
+                    env={
+                        'NO_STOP_MESSAGE':'1', # For Fortran test-suite
+                    },
+                    testsuite_flags=[
+                        '--cppflags', '-mcpu=neoverse-v2 -mllvm -scalable-vectorization=preferred -mllvm -treat-scalable-fixed-error-as-warning=false -O3',
+                        '--threads=32', '--build-threads=32'],
```
maxim-kuvyrkov, line 583: The current instances have 48 cores and 4GB RAM per core, so we should be OK with max parallelism of 48.  Same comment below for 2stage builder.

maxim-kuvyrkov (approved):
Other than the parallelism comment -- looks great.  Thanks!

buildbot/osuosl/master/config/builders.py
```diff
@@ -586,8 +586,7 @@
                         "-DCMAKE_CXX_FLAGS='-mcpu=neoverse-v2'",
                         "-DLLVM_ENABLE_LLD=True",
                         "-DMLIR_INCLUDE_INTEGRATION_TESTS=True",
-                        "-DMLIR_RUN_ARM_SVE_TESTS=True",
-                        "-DLLVM_LIT_ARGS='-v'"])},
```
maxim-kuvyrkov: [Just checking] Will you remove "-v" from the existing configurations in a separate PR?

DavidSpickett: ....I will now that you have brought it up :)

buildbot/osuosl/master/config/builders.py
DavidSpickett: Done.

buildbot/osuosl/master/config/builders.py
DavidSpickett: https://github.com/llvm/llvm-zorg/pull/322

omjavaid:
This looks good to me overall. just one minor nit inline.

omjavaid (approved):

DavidSpickett:
CI failure is due to an unrelated change, merging.

PR #322, opened 2024-11-28
Remove explicit LLVM_LIT_ARGS='-v' from Linaro builders

DavidSpickett:
As this is added by the build factories anyway.

Arm lldb has a thread limit so I've kept that in.

(this because the machine it runs on actually has a lot of headroom, but not all the time, so assuming you can always use every thread makes the results flaky)


asb (approved):
LGTM

maxim-kuvyrkov (approved):
Thanks!

PR #323, opened 2024-11-28
Fix local testing mode by skipping downstream-only config setting

asb:
Ideally we'd better replicate the the deployed config, but this hopefully moves things forward. Note: as the GitHub Actions pre-commit testing relies on "local testing mode", this PR fixes that too.

@gkistanova the requirements.txt just references a local checkout for the buildbot version being used - the other libs are all 3.11.7 so that's what I use for the GitHub actions and the guidance for "local testing mode". Is there a public repo of the buildbot fork that is deployed?

### DIFF
--- a/buildbot/osuosl/master/master.cfg
+++ b/buildbot/osuosl/master/master.cfg
@@ -39,7 +39,8 @@ c['workers'] = config.workers.get_all()
 
 c['protocols'] = {'pb': {'port': "tcp:9990:interface=127.0.0.1" if test_mode else 9990}}
 
-c['ignoreOfflineWorkersTimeout'] = 30 # minutes.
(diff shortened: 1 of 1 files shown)


nickdesaulniers (approved):
Ah! This is a downstream only change to buildbot! No wonder why I couldn't find any documentation about this. It's blocking my presubmit testing of https://github.com/llvm/llvm-zorg/pull/325, so +1.

asb:
Thanks for the review, merging now to hopefully avoid more wasted time due to the caused confusion.

Once you get a chance @gkistanova it would be great if you could say more about the downstream buildbot patches and if there's a publicly accessible repo (perhaps it could be created under github.com/llvm ?). Thanks!

PR #324, opened 2024-11-29
[CI] Upstream terraform infrastructure

boomanaiden154:
This patch contains the terraform configuration for the premerge infrastructure. Actually applying changes requires someone from within Google with access to the GCP project to apply the changes, but the entire infrastructure is described within the TF in this patch.

Co-authored-by: Nathan Gauër <nathangauer@google.com>

### DIFF
--- a/premerge/README.md
+++ b/premerge/README.md
@@ -1,0 +1,31 @@
+# Premerge Infrastructure
+
+This folder contains the terraform configuration files that define the GCP
+resources used to run the premerge checks. Currently, only Google employees
+with access to the GCP project where these checks are hosted are able to apply
+changes. Pull requests from anyone are still welcome.
+
+## Setup
+
+- install terraform (https://developer.hashicorp.com/terraform/install?product_intent=terraform)
+- get the GCP tokens: `gcloud auth application-default login`
+- initialize terraform: `terraform init`
+
+To apply any changes to the cluster:
+- setup the cluster: `terraform apply`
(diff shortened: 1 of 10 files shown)


boomanaiden154:
https://github.com/llvm/llvm-project/pull/117397, but in https://github.com/llvm/llvm-zorg.

Keenuts (approved):
LGTM from the GCP/cluster point of view.
Letting the others approve/review for repo organization, etc

premerge/README.md
```diff
@@ -0,0 +1,31 @@
...
+terraform apply -target google_container_node_pool.llvm_premerge_linux_service
+terraform apply -target google_container_node_pool.llvm_premerge_linux
+terraform apply -target google_container_node_pool.llvm_premerge_windows
+terraform apply
+```
+
+Setting the cluster up for the first time is more involved as there are certain
+resources where terraform is unable to handle explicit dependencies. This means
+that we have to set up the GKE cluster before we setup any of the Kubernetes
+resources as otherwise the Terraform Kubernetes provider will error out.
```
joker-eph, line 31: Thanks for providing a documentation!

joker-eph:
That LGTM in general, I'm not a terraform expert though so I wouldn't be able to to spot any subtle issue there!

premerge/README.md
boomanaiden154: Yep! There should be more to follow too.

PR #332, opened 2024-12-04
libc: make new bootstrap builder a dup of runtimes

nickdesaulniers:
Before renaming the existing debian-dbg-runtimes-build to
debian-dbg-bootstrap-build, make sure that the new name will be recognized so
that there's no downtime of CI coverage due to renaming the buildbot.

Also remove llvm from the explicit project list; that's not necessary (or may have
been for old hdrgen which has been removed).

Link: https://github.com/llvm/llvm-zorg/pull/325#discussion_r1870167521

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2237,10 +2237,10 @@ all += [
                     depends_on_projects=['llvm', 'libc', 'clang', 'clang-tools-extra'],
                     extra_args=['--debug', '--asan'])},
 
-    {'name' : "libc-x86_64-debian-dbg-runtimes-build",
+    {'name' : "libc-x86_64-debian-dbg-bootstrap-build",
     'tags'  : ["libc"],
     'workernames' : ["libc-x86_64-debian"],
-    'builddir': "libc-x86_64-debian-dbg-runtimes-build",
+    'builddir': "libc-x86_64-debian-dbg-bootstrap-build",
     'factory' : AnnotatedBuilder.getAnnotatedBuildFactory(
                     script="libc-linux.py",
                     depends_on_projects=['llvm', 'libc', 'clang', 'clang-tools-extra'],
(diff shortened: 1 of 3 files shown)


nickdesaulniers:
Marking as draft. I can copy the existing sources to a newly named dir on the bot, then update 2 more files to just create the new worker (IIUC).

nickdesaulniers:
Ok, I broke this up into distinct patches which are perhaps easier to review individually (will squash when merging though).

This PR should convert the existing buildbot IIUC.

nickdesaulniers:
Might want to drop that last commit. Reading the zorg sources trying to make sense of what `depends_on_projects` implies.

nickdesaulniers:
ping for review

lntue (approved):

nickdesaulniers:
Hi @gkistanova . I was suspecting that this PR would have changed the "[libc-x86_64-debian-dbg-runtimes-build](https://lab.llvm.org/buildbot/#/builders/78)" builder to be "libc-x86_64-debian-dbg-bootstrap-build", but a day later the builder still has not changed. Is there something on your end that needs to explicitly be reset? Or is there something else I need to do?

I've read through https://llvm.org/docs/HowToAddABuilder.html#steps-to-add-builder-to-llvm-buildbot, but I don't see what I'm missing.

DavidSpickett:
I think you are at:
> 9. Send the buildbot-worker access name and the access password directly to [Galina Kistanova](mailto:gkistanova%40gmail.com), and wait until she lets you know that your changes are applied and buildmaster is reconfigured.

But just the latter part, given that the worker is already setup.

Changes to llvm-zorg are not automatically applied, at least for the "normal" master which is the one your specific bot is connecting to. Staging might be updated more often, but that's just my impression.

When the changes are applied, this bot name will likely still be in the results but you'll see new builds under the new name.

DavidSpickett:
There is certainly room there for an abridged version of adding a worker as "making a change".

PR #335, opened 2024-12-06
Set Fortran flags on AArch64 SVE/SVE2 builders

luporl:
When running the llvm-test-suite, pass the C++ flags to the
Fortran compiler too, to check that it behaves correctly
with them.


DavidSpickett (approved):
LGTM, thanks for testing this!

PR #336, opened 2024-12-09
[AMDGPU][libc] Disallow collapsing of requests

jplehr:
We have seen on this bot in particular issues with picking up a breaking change and some NFC change. This leads to the bot not commenting on the PR.


Meinersbur:
Are you sure a single worker able to keep up with the workload? I've seen a builder lagging behind 3 months because it was slower processing commits than there were new onese.

jplehr:
I would like to give it a shot. I monitor those workers relatively closely and we can revert this change or move the bot to another machine when we see that it starts lagging (too much).

Meinersbur (approved):

jplehr:
@gkistanova this bot is in production, so I believe we need a master restart. This is _not_ urgent. Let me know if you need anything from our end.

PR #337, opened 2024-12-11
Extend avx512 win job with Intel oneAPI

DoyleLi:
The reason for the change is to mitigate the failure during compiling when using Intel oneAPI product. 
The check from buildbot will prevent the commit to be merged if the job failed.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -910,6 +910,8 @@ all = [
                     extra_cmake_args=[
                         "-DCMAKE_C_FLAGS='-march=cascadelake'",
                         "-DCMAKE_CXX_FLAGS='-march=cascadelake'",
+                        "-DCMAKE_C_COMPILER='icx.EXE'",
+                        "-DMAKE_CXX_COMPILER='icx.EXE'",
                         "-DLLVM_ENABLE_RUNTIMES=compiler-rt",
                         "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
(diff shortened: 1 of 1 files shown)


gkistanova (changes requested):
Could you update the description with the rationale of this change, please?

DoyleLi:
> Could you update the description with the rationale of this change, please?

Sure, thanks for the note! I have updated in the PR.

buildbot/osuosl/master/config/builders.py
```diff
@@ -910,6 +910,8 @@
                     extra_cmake_args=[
                         "-DCMAKE_C_FLAGS='-march=cascadelake'",
                         "-DCMAKE_CXX_FLAGS='-march=cascadelake'",
+                        "-DCMAKE_C_COMPILER='icx.EXE'",
```
gkistanova, line 913: You do not need quotation marks around the executable name here and in the line below. Quotation marks wrapping could be tricky on Windows, better to avoid unneeded all together.

gkistanova (approved):
Thanks, @DoyleLi!

LGTM with a small nit.
Please see my comment inline.

It is Ok to commit after fixing this.

buildbot/osuosl/master/config/builders.py
DoyleLi: Thank you so much, @gkistanova.

gkistanova (approved):
LGTM.

DoyleLi:
Hello @gkistanova  Could you please help to merge the PR? Thanks!

PR #338, opened 2024-12-11
[Offload] Adds buildbot for CMake cache file

jplehr:
The buildbot uses the `offload/cmake/cache/AMDGPUBot.cmake` file to create the build config. It uses an annotated builder for that and only performs the build step as most of the post-commit issues are build problems and not test errors.

A lot of inspiration was taken from the `libc` annotated builder script in the zorg repository.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1932,35 +1932,15 @@ all += [
                         add_openmp_lit_args=["--time-tests", "--timeout 100"],
                         )},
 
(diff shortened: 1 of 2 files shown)


jplehr:
I set this into draft mode, since we are not fully there on the machine side of things. However, I tested this locally (thank you for making this mode possible!), and it seems to do what I want it to do.

The midterm goal is to move more of our testing towards using the mechanism here, i.e., a CMake cache file for configuration. In addition, I'm preparing to make the docker recipes public that are the base of our buildbots.
This aims to enable people to reproduce what the buildbot does locally.

PR #340, opened 2024-12-12
[AMDGPU][libc] Re-enable batching to keep up

jplehr:
The current config is unable to keep up with the load on that particular bot, meaning that the build queue filled-up and was at several hours. This reverts the introduced "neverCollapse" and does allow collapsing of build requests.


jhuber6 (approved):

DavidSpickett:
FYI, Linaro's longer running bots also had this issue even with collapsing enabled. If this happens to you, there is a further step you can take - https://github.com/llvm/llvm-zorg/pull/266.

I think you would need to add `clean` to `OpenMPBuilder` and then it should "just work". If your build steps are already being ccached, it's not much extra overhead.

PR #344, opened 2024-12-17
[Offload] Adds buildbot for CMake cache file

jplehr:
The buildbot uses the offload/cmake/cache/AMDGPUBot.cmake file to create the build config. It uses an annotated builder for that and only performs the build step as most of the post-commit issues are build problems and not test errors.

### DIFF
--- a/zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py
+++ b/zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py
@@ -1,0 +1,56 @@
+#!/usr/bin/python
+
+import argparse
+import os
+import subprocess
+import sys
+import traceback
+import util
+from contextlib import contextmanager
+
+
+def main(argv):
+    source_dir = os.path.join("..", "llvm-project")
+    offload_base_dir = os.path.join(source_dir, "offload")
+    of_cmake_cache_base_dir = os.path.join(offload_base_dir, "cmake/caches")
+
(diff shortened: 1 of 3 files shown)


zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py
```diff
@@ -0,0 +1,56 @@
...
+    of_cmake_cache_base_dir = os.path.join(offload_base_dir, "cmake/caches")
+
+    with step("cmake", halt_on_fail=True):
+        # TODO make the name of the cache file an argument to the script.
+        cmake_cache_file = os.path.join(of_cmake_cache_base_dir, "AMDGPUBot.cmake")
+
+        # Use Ninja as the generator.
+        # The other important settings alrady come from the CMake CMake
+        # cache file inside LLVM
+        cmake_args = ["-GNinja", "-C %s" % cmake_cache_file]
```
jhuber6, line 24: Can we not just put this in the extra Cmake args?

jplehr: I'm not sure whether they exist on the `AnnotatedBuilder` that this is using.

To be clear, this is not using the OpenMPBuilder buildbot that we have based the other buildbots on.

jplehr:
I removed the unrelated new-line and blank changes.

gkistanova (approved):
LGTM.

PR #347, opened 2024-12-25
Change linker from lld to gold for builder.

dyung:
When I originally adder this builder, I'm not sure why I changed the linker to be lld, it should be gold to match our previous instance. This change fixes that.


gkistanova (approved):
LGTM.

PR #349, opened 2025-01-03
[AMDGPU][Offload] Adds RHEL 8 CMake cache file builder

jplehr:


jplehr:
Does the build directory be unique across the fleet of builders?

DavidSpickett:
This is from a worker we have that is used for 4 different builders:
```
/home/tcwg-buildbot/worker# ls
buildbot.tac  clang-aarch64-sve-vla  clang-aarch64-sve-vla-2stage  clang-aarch64-sve-vls  clang-aarch64-sve-vls-2stage  <...>
```
There is only one builder that builds in a folder called `clang-aarch64-sve-vla`, but it does so across many workers.

If the build folder was more generic like "build-llvm" then in theory as long as two differently configured builders never landed on the same worker, there would be no problem. It's easy to name the folder uniquely though, so I'd do it anyway.

And it provides you a clue if, for example, your disk fills up because of one specific builder. With unique build folder names you can tell which one did it from looking at the worker machine.

buildbot/osuosl/master/config/builders.py
```diff
@@ -1962,6 +1962,16 @@
                     script_interpreter=None
                 )},
 
+    {'name' : "amdgpu-offload-rhel-8-cmake-build-only",
+    'tags'  : ["amdgpu", "offload", "openmp"],
+    'workernames' : ["rocm-docker-rhel-8"],
+    'builddir': "bbot-build",
```
DavidSpickett, line 1968: Personally I'd put "amdgpu-offload-rhel-8-cmake-build-only" here to match the builder name, but what you have shouldn't be a problem given that the two builders will never build on the same worker.

DavidSpickett:
If the distinction between "builder" and "worker" isn't clear (it never has been :) ), slide 5 of https://llvm.org/devmtg/2022-11/slides/TechTalk5-WhatDoesItTakeToRunLLVMBuildbots.pdf tries to clarify.

More often than not in LLVM's bots we have a specific worker (or set of) for each builder, so it can seem like builder and worker are the same thing, but it does not have to be done that way.

In your case you need a specific OS installed so a specific worker is a good way to ensure that.

DavidSpickett:
Actually Buildbot does seem to enforce unique build dirs, from the CI result:
```
Configuration Errors:
  duplicate builder builddir 'bbot-build'
```
Perhaps a bit pessimistic but I can see why they'd do that.

jplehr:
Thank you for that elaborate answer. I did understand that there is a difference, however, so far, we have always had a 1:1 mapping.
With that question I was mostly referring to the CI result that seems to indicate unique directories.

jplehr:
I put in a unique builddir and a custom `depends_on_projects` to match what we have in the other offload builders.

gkistanova (approved):
LGTM.

PR #352, opened 2025-01-07
[Offload][AMDGPU] RHEL9.4 CMake Cache builder

jplehr:
Dockerized RHEL 9.4 builder that uses the Offload CMake Cache file AMDGPUBot.cmake as build config.


gkistanova (approved):
LGTM.

PR #354, opened 2025-01-10
Compile for Neoverse N1 on Linaro's Armv8 bots that are on Mt Jade hardware

DavidSpickett:
This reflects the host hardware, rather than the limited subset of features on the Cortex-A57.

```
$ lscpu
Architecture:            aarch64
  CPU op-mode(s):        32-bit, 64-bit
<...>
Vendor ID:               ARM
  Model name:            Neoverse-N1
<...>
    Flags:               fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid
                         asimdrdm lrcpc dcpop asimddp
```

I have also confirmed that the Neoverse V1 (Note: V not N) that is used for Graviton is a superset of these features.

https://github.com/aws/aws-graviton-getting-started/blob/main/c-c%2B%2B.md#enabling-arm-architecture-specific-features

So if we have to move these builds to Graviton 3 at short notice, this will still work (Graviton 4 also ok for AArch64, but it drops the AArch32 mode).

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -393,11 +393,11 @@ all = [
                 useTwoStage=True,
                 runTestSuite=True,
                 testsuite_flags=[
-                    '--cppflags', '-mcpu=cortex-a57 -fuse-ld=lld',
+                    '--cppflags', '-mcpu=neoverse-n1 -fuse-ld=lld',
                     '--threads=32', '--build-threads=32'],
                 extra_cmake_args=[
-                    "-DCMAKE_C_FLAGS='-mcpu=cortex-a57'",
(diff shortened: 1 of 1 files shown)


luporl (approved):
LGTM.

omjavaid (commented):
Shouldn't we also update clang-armv8-lld-2stage builder ??

DavidSpickett:
> Shouldn't we also update clang-armv8-lld-2stage builder ??

Done.

There are v7 builds running on Neoverse-N1 but it makes sense to leave them targeting CPUs that are actually v7, even if they are going to ultimately run on Armv8.

DavidSpickett:
@omjavaid Good to go in with the armv8 changes?

omjavaid (approved):
Yes LGTM Thanks

PR #358, opened 2025-01-16
[RISCV] Move rva23 evl builder over to cross-compile and execute under qemu-system setup

asb:
This also moves to
`prefer-predicate-over-epilogue=predicate-else-scalar-epilogue`. As the buildbot is on staging only and already has some failures, I think it's ok to make both changes at once rather than carefully staging it.


buildbot/osuosl/master/config/builders.py
```diff
@@ -3353,9 +3356,27 @@
...
+                    "set(CMAKE_C_FLAGS_INIT '-march=rva23u64 -mllvm -force-tail-folding-style=data-with-evl -mllvm -prefer-predicate-over-epilogue=predicate-else-scalar-epilogue')",
+                    "set(CMAKE_CXX_FLAGS_INIT '-march=rva23u64 -mllvm -force-tail-folding-style=data-with-evl -mllvm -prefer-predicate-over-epilogue=predicate-else-scalar-epilogue')",
+                    "set(CMAKE_LINKER_TYPE LLD)",
+                    "set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)",
+                    "set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)",
+                    "set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)",
+                    "set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)"],
+                env={
+                    "BB_IMG_DIR": util.Interpolate("%(prop:builddir)s/.."),
+                    "BB_QEMU_CPU": "zba=true,zbb=true,zbc=false,zbs=true,zfhmin=true,v=true,vext_spec=v1.0,zkt=true,zvfhmin=true,zvbb=true,zvkt=true,zihintntl=true,zicond=true,zcb=true,zfa=true,zawrs=true,rvv_ta_all_1s=true,rvv_ma_all_1s=true,rvv_vl_half_avl=true",
```
lukel97, line 3377: Do we also need to enable zimop +zcmop? Can clang emit those currently?

lukel97 (approved):
LGTM thanks, just one question about QEMU_CPU

buildbot/osuosl/master/config/builders.py
asb: I don't believe we will, but I'll enable them for completeness - thanks. They weren't supported at the time I composed that command line, but they were added in qemu 9.1.0.

Thanks!

asb: A patch is under review for rva23 support in QEMU, so I'll additionally add a TODO here to switch over to that shorthand once it's merged and in a released qemu version https://lore.kernel.org/qemu-devel/20250115184316.2344583-1-dbarboza@ventanamicro.com/

PR #360, opened 2025-01-18
[CI] Switch away from kubernetes executor mode

boomanaiden154:
We are running into reliability problems with the kubernetes executor mode, hypothesized currently to be due to a complex interaction with konnectivity-agent pods dying and subsequently killing in-process execs through the k8s control plane. The plan is to switch back to the in-container executor mode for now while we sort out the issue upstream with the konnectivity developers.

This is related to #362.


boomanaiden154:
`ci-ubuntu-22.04-agent` currently does not exist yet. I need to put up another PR against the monorepo's container build job to generate it. I don't want to bundle the runner in with the normal container as it bloats the size by 40-50%. This can land once that has landed.

boomanaiden154:
https://github.com/llvm/llvm-project/pkgs/container/ci-ubuntu-22.04-agent exists now, so we should be good to go to land this. 

Once this lands I'll `terraform apply` and land https://github.com/llvm/llvm-project/pull/123483 at the same time, and it should be a relatively smooth transition.

Keenuts (approved):
LGTM, but as the other PR: please link the description to an llvm-zorg issue describing the reliability issues symptoms and the switch, including konnectivity issue link.

PR #363, opened 2025-01-22
Collapse requests for the flang AIX buildbot

madanial0:
allow collapseRequests for flang AIX bot: https://lab.llvm.org/buildbot/#/builders/201, due to a long build queue. Most aix bots already allow collapseRequests so this would be consistent with existing behaviour.


DavidSpickett:
FYI you can clear that queue - https://llvm.org/docs/HowToAddABuilder.html#managing-a-worker-from-the-web-interface

This helps if it's a temporary buildup. I suspect request collapsing happens as the request enters the queue, so the old ones need clearing out.

lei137 (approved):
LGTM
Thx

gkistanova (approved):
LGTM.

PR #364, opened 2025-01-23
[CI] Add shebang to shim wrapper scripts

justice-adams-apple:
After the update of the Cmake binary on our CI fleet, the way Cmake handles internal processes has changed and we need to explicitly set the shebang on our wrapper shims to avoid 
```
    Error running link command: Unknown system error -8
```


etcwilde (approved):
For documentation purposes, CMake 3.28 (IIRC, though it got reverted a few times), switched from a custom solution for managing process execution to using libuv. This results in the script that is doing the execution being handled differently in `cmake -E cmake_linker_script <wrapper_script>`. Instead of being treated as a shell script, it's being treated as a raw executable binary. This doesn't work since it's a text file.

Adding the shebang or calling `bash <script>` in the wrapper script will fix it.
LGTM.

PR #374, opened 2025-02-09
[RISCV] Move clang-riscv-rva23-evl-vec-2stage to annotated builder

asb:
The intent is for the rest of the RISE RISC-V builders to follow, but this starts the process.

Previously the builders used the framework provided by ClangBuilder.py. As the build has required more customisation (with more to come - such as building the llvm-test-suite with the stage1 compiler which isn't support by ClangBuilder) this is requiring more and more changes to increase configurability of ClangBuilder, but this makes the logic harder to follow and I think we've reached the limit of what's reasonable.

The AnnotatedBuilder infrastructure allows the build logic to just be put in a shell script. The latest version of llvm-zorg is checked out and the script is executed. This has the additional advantage of decoupling script updates from needing to wait for buildmaster redeploys. With that in mind, I've avoided putting any configuration through options or environment variables defined by the builder instantiation in builders.py so there's no need to reason about cases where e.g. a script update was committed but the buildmaster may or may not have been updated with a new builders.py.

rise-riscv-build.sh is written with the philosophy that "A little copying is better than a little dependency." We could look to share more logic, but we'd end up with the same configurability issue as ClangBuilder. Instead, the script provides only the configurability needed for this set of builders.

A future improvement would be to make rise-riscv-build.sh more friendly to running outside of the buildbot environment (e.g. for local testing). This is left to future work, as it's not a regression vs the current approach.

This change has been tested locally along the lines of the instructions I wrote in
<https://llvm.org/docs/HowToAddABuilder.html#testing-a-builder-config-locally>. Something like the following is sufficient to use a local downstream testing llvm-zorg branch for the annotated builder scripts temporarily:

```
--- a/zorg/buildbot/builders/AnnotatedBuilder.py
+++ b/zorg/buildbot/builders/AnnotatedBuilder.py
@@ -77,6 +77,8 @@ def getAnnotatedBuildFactory(
     # Check out zorg so we can run the annotator scripts.
     f.addGetSourcecodeForProject(
         name='update-annotated-scripts',
+        repourl='file:///home/asb/llvm-zorg',
+        branch='synced-worktree',
         project='zorg',
         src_dir='llvm-zorg',
         alwaysUseLatest=True)
```

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -3362,46 +3362,11 @@ all += [
     {'name' : "clang-riscv-rva23-evl-vec-2stage",
     'workernames' : ["rise-clang-riscv-rva23-evl-vec-2stage"],
     'builddir':"clang-riscv-rva23-evl-vec-2stage",
-    'factory' : ClangBuilder.getClangCMakeBuildFactory(
-                clean=True,
-                useTwoStage=True,
-                runTestSuite=False,
-                testStage1=False,
-                checkout_compiler_rt=False,
-                checkout_zorg=True,
-                extra_cmake_args=[
-                    "-DCMAKE_C_COMPILER=clang",
-                    "-DCMAKE_CXX_COMPILER=clang++",
-                    "-DLLVM_ENABLE_LLD=True",
-                    "-DLLVM_TARGETS_TO_BUILD=RISCV",
-                    "-DCMAKE_C_COMPILER_LAUNCHER=ccache",
-                    "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache"],
-                extra_stage2_cmake_args=[
-                    util.Interpolate("-DLLVM_NATIVE_TOOL_DIR=%(prop:builddir)s/stage1.install/bin"),
-                    "-DLLVM_BUILD_TESTS=True",
-                    "-DPython3_EXECUTABLE=/usr/bin/python3",
-                    "-DLLVM_HOST_TRIPLE=riscv64-linux-gnu",
-                    util.Interpolate("-DLLVM_EXTERNAL_LIT=%(prop:builddir)s/llvm-zorg/buildbot/riscv-rise/lit-on-qemu")],
-                stage2_toolchain_options=[
-                    "set(CMAKE_SYSTEM_NAME Linux)",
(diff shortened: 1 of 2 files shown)


lukel97:
This seems reasonable to me

> As the build has required more customisation (with more to come - such as building the llvm-test-suite with the stage1 compiler which isn't support by ClangBuilder)

Is this specifically with a two stage configuration? I thought that `clang-aarch64-sve-vla` was able to build llvm-test-suite with the stage1 compiler, and then it has a separate `clang-aarch64-sve-vla-2stage` builder to build it it with the stage2?

Is the plan to have `clang-riscv-rva23-evl-vec-2stage` do both a stage1 llvm-test-suite build and build a stage2 compiler?

lukel97:
> Building and running the test suite is very fast, so as noted in that PR it would be interesting to move it to run the test suite immediately after the stage1 build so if it's broken we get that answer quickly (though would likely want to proceed to build + test stage2 even if there was a llvm-test-suite failure).

If we wanted to test both stage1 and stage2, would it be easiest to split this out into two separate builders?

I'm not sure if this would result in an explosion of builders though. 

AArch64 has `clang-aarch64-{sve-vla,sve-vls,sve2-vla}{,-2stage}`, so 6 in total. 

For RISC-V we would end up with `clang-riscv-{rva20,rva23,rva23-evl,rva23-mrvv}{,-2stage}`, so 8 in total. I noticed we have dedicated workers for each builder whereas AArch64 is able to pool them, so maybe this isn't practical. 

In any case I agree it's not worth trying to fight the ClangBuilder.py logic, so if this patch makes life easier I'm all for it

lukel97 (approved):

asb:
> If we wanted to test both stage1 and stage2, would it be easiest to split this out into two separate builders?

It's definitely an option to consider, and much more feasible now workers no longer run "natively" (which meant before an rva20 and rva23 worker weren't interchangeable, as they were different qemu-system environments). Though if we just did stage1 and llvm-test-suite then the failure emails may be somewhat confusing due to not including LLVM unit test failures (i.e. a problem in llvm-test-suite may act as a red herring vs a much more obvious unit test failure). If we did stage1 + LLVM unit tests + llvm-test-suite then that's burning a fair bit more CPU for testing that's already covered by other builders.

After this series of patches the plan is to look at the best way of trading off the total CPU power we have vs testing coverage vs feedback time.

PR #375, opened 2025-02-09
[RISCV] Move clang-riscv-rva20-2stage to annotated builder

asb:
Followup to #374 (and stacks on top of it), but this time converting the RVA20 builder.


lukel97 (approved):
LGTM

PR #376, opened 2025-02-10
[RISCV] Add llvm-test-suite to RISC-V annotated builders

asb:
As noted in the inline TODO, it may be worth later changing this so the
test suite is run immediately after stage1 as stage1 + llvm-test-suite
can build and run very quickly, provided rapid feedback.

Although the lit-on-qemu mechanism could be used to run the test suite
under qemu-system, we instead run the test suite under qemu-user. This
has the advantage of simplicity - the suite runs well in this mode and
we don't have the same concerns about testing things that may not work
under qemu-user (such as LLDB or sanitizers) as for the main
llvm-project tests.

### DIFF
--- a/zorg/buildbot/builders/annotated/rise-riscv-build.sh
+++ b/zorg/buildbot/builders/annotated/rise-riscv-build.sh
@@ -129,6 +129,7 @@ export QEMU_LD_PREFIX="$(pwd)/../rvsysroot"
 export QEMU_CPU="$BB_QEMU_CPU"
 cmake -G Ninja \
   --toolchain=$(pwd)/stage1-toolchain.cmake \
+  -DCMAKE_BUILD_TYPE=Release \
   -DTEST_SUITE_LIT=$(pwd)/stage1/bin/llvm-lit \
   -DTEST_SUITE_LIT_FLAGS=-v \
   -DTEST_SUITE_COLLECT_CODE_SIZE=OFF \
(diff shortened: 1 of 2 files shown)


zorg/buildbot/builders/annotated/rise-riscv-build.sh
```diff
@@ -112,3 +112,33 @@ cmake --build stage2
...
+build_step "Updating llvm-test-suite repo"
+git -C llvm-test-suite fetch origin
+git -C llvm-test-suite reset --hard origin/main
+
+build_step "llvm-test-suite cmake"
+export QEMU_LD_PREFIX="$(pwd)/../rvsysroot"
+export QEMU_CPU="$BB_QEMU_CPU"
+cmake -G Ninja \
+  --toolchain=$(pwd)/stage1-toolchain.cmake \
+  -DTEST_SUITE_LIT=$(pwd)/stage1/bin/llvm-lit \
```
lukel97, line 132: Just want to check, if we don't specify `-C cmake/caches/O3.cmake` etc., does it default to building the release configuration? I think it's better if we can leave this as implicit since it doesn't look like ClangBuilder.py specifies it either (but also runs it via lnt: https://github.com/llvm/llvm-zorg/blob/3c627f0fa9da5566c88600b45e8293821a9fde2e/zorg/buildbot/builders/ClangBuilder.py#L560-L579)

lukel97 (approved):
LGTM

zorg/buildbot/builders/annotated/rise-riscv-build.sh
asb: It does default to CMAKE_BUILD_TYPE=Release which defaults to `-O3 -DNDEBUG`.

I think it's probably good to avoid people having to check the default behaviour when reading the build script, so I'll go ahead and add an explicit `-DCMAKE_BUILD_TYPE=Release`.

PR #379, opened 2025-02-15
Move Grafana Dashboard Definitions into TF

boomanaiden154:
This patch moves the grafana dashboard definition into TF and creates a new JSON file defining the dashboard (pulled from the web interface). This should make the setup a bit more transparent to the community and at least give an interface for others to hack on even if it is less convenient than modifying the dashboard through the web UI.

### DIFF
--- a/premerge/main.tf
+++ b/premerge/main.tf
@@ -4,6 +4,10 @@ terraform {
       source  = "hashicorp/google"
       version = "6.17.0"
     }
+    grafana = {
+      source  = "grafana/grafana"
+      version = "3.19.0"
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


lnihlen:
So what's the source of truth now? If we change the dash do we need to update this?

boomanaiden154:
> So what's the source of truth now? If we change the dash do we need to update this?

The idea is that this would become the source of truth. Updating the dashboard could either be done by updating through the web UI and copying the JSON into the attached file, or by directly modifying the JSON and then running `terraform apply`.

Not entirely sure if this is what we want to end up doing though. I think it increases transparency to those without direct access to the Grafana instance to edit things and enables others to implement simple fixes, but validating any non-trivial change will be difficult. And it slightly complicates the existing process as we need to make sure the in-tree JSON file stays up to date when making changes through the web UI.

Keenuts:
I like the idea, but I'm not sure how usable this will be in practice:
 - it only covers the dashboards, not the alerting/monitoring around it.
 - every time I had to do a dashboard change, it was mostly try&repeat, as there is no easy way to test a change.

So the flow would mostly be:
 - try & repeat in the grafana UI (edit permission required)
 - export a JSON file
 - add it to this repo, and re-deploy
This adds more work, but doesn't bring IMO a clear benefit (grafana does had edit history).
In addition, I don't see this as a real benefit for an external contributor:
 - dashboard metrics definitions can be observed from grafana
 - if an edit is requested, they wouldn't be able to test it, and we also we wouldn't be able to properly test it, except by merging this, observing the change, and possibly reverting/re-doing the flow mentionned above.

boomanaiden154:
Yeah, I came to somewhat similar conclusions after getting it implemented. Probably a better idea in theory than practice. Thanks for the feedback.

PR #380, opened 2025-02-17
[AMDGPU] Disallow batching in build-only bots

jplehr:
These should be fast enough to keep up with collapsing disabled.


jplehr:
These bots are in production. Let me know if we need to do anything after this change landed in order for the bots to update.

jhuber6 (approved):

PR #384, opened 2025-02-18
[AMDGPU][Libc] Four parallel GPU processes

jplehr:
Gentle increase to the number of GPU tests run in parallel on the device. Tested this locally, this should be a reasonable balance between speed up and system load.


jhuber6 (approved):
I've had the test fail with as few as two given a few thousand runs, but we'll see.

jplehr:
Yeah, since this bot is still in staging and running latest ROCm, let's see what happens.

PR #333, opened 2024-12-06
Enable Flang-RT for all Flang builders.

Meinersbur:
Modify all build configurations that build Flang to also build Flang-RT. Flang-RT is the runtime component of Flang that is refactored into its own runtime project in https://github.com/llvm/llvm-project/pull/110217. 

The following classes of builder are affected:

 * Builders with `LLVM_ENABLE_PROJECTS=flang` and running `check-flang`: Adding `LLVM_ENABLE_RUNTIMES=flang-rt` and `check-flang-rt`. These builders would continue to be green after https://github.com/llvm/llvm-project/pull/110217 even without this PR, just testing less.

 * Clang builders with `checkout_flang=True`: Automatically add `flang-rt` to `depends_on_projects` which also adds `LLVM_ENABLE_PROJECTS` with `enable_runtimes="auto".` These builders would continue to be green after https://github.com/llvm/llvm-project/pull/110217 even without this PR, just testing less.

 * Clang builders with `checkout_flang=True` and `runTestSuite=True`: Running the the test-suite with Flang requires Flang-RT, so these builders would start failing with https://github.com/llvm/llvm-project/pull/110217 without this PR.

 * linaro-flang-aarch64-out-of-tree: Has its own build factory that would require larger changes to also check/run Flang-RT. Should stay green even after https://github.com/llvm/llvm-project/pull/110217.

 * flang-runtime-cuda-gcc: Only compiles but does not run check. Used to be a `flang/runtime` out-of-tree build. Runtime-only builds of Flang-RT needs the path to `flang`. I do not know where/whether this can be found on the workers so changing it to a bootstrap build. More important to keep the buildbots green.

 * flang-runtime-cuda-clang: Staged build of first compiling LLVM+Clang, then Flang out-of-tree. With this PR also building Flang in-tree and Flang-RT as runtime-only build. Building both Flang and Flang-RT out-of-tree would require another stage, but "linaro-flang-aarch64-out-of-tree" is already testing the Flang out-of-tree build.

PR #387, opened 2025-02-21
LLVM_ENABLE_RUNTIMES=flang-rt for UnifiedTreeBuilder builders

Meinersbur:
Add `depends_on_projects=['flang-rt']` and `checks=['check-flang-rt']` to Linaro's builders that are based on UnifiedTreeBuilder. This prepares the removal of the "projects" build of the flang runtime in https://github.com/llvm/llvm-project/pull/124126.

Split off from #333

Affected builders:
 * flang-aarch64-dylib
 * flang-aarch64-sharedlibs
 * flang-aarch64-debug-reverse-iteration
 * flang-aarch64-libcxx
 * flang-aarch64-release
 * flang-aarch64-rel-assert
 * flang-aarch64-latest-gcc

Affected workers:
 * linaro-flang-aarch64-dylib
 * linaro-flang-aarch64-sharedlibs
 * linaro-flang-aarch64-debug-reverse-iteration
 * linaro-flang-aarch64-libcxx
 * linaro-flang-aarch64-release
 * linaro-flang-aarch64-rel-assert
 * linaro-flang-aarch64-latest-gcc

I could locally confirm that `flang-aarch64-dylib` works on a x86_64 host (since I don't have an aarch machine other than a Raspberry Pi readily available)

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2353,8 +2353,8 @@ all += [
     'builddir': "flang-aarch64-dylib",
     'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
                     clean=True,
-                    checks=['check-flang'],
-                    depends_on_projects=['llvm','mlir','clang','flang','openmp'],
+                    checks=['check-flang','check-flang-rt'],
+                    depends_on_projects=['llvm','mlir','clang','flang','flang-rt','openmp'],
                     extra_configure_args=[
                         "-DLLVM_TARGETS_TO_BUILD=AArch64",
                         "-DLLVM_BUILD_LLVM_DYLIB=ON",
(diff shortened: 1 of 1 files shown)


ceseo:
> LGTM.
> 
> @ceseo, you've already confirmed that LLVM_ENABLE_RUNTIMES=flang-rt and check-flang-rt work on AArch64/Linux, right?

Yes, both stage1 and stage2.

ceseo (approved):

Meinersbur:
After a buildbot master restart (only) [flang-aarch64-libcxx](https://lab.llvm.org/buildbot/#/builders/89) stopped working. The error message
```
/usr/bin/ld: flang-rt/unittests/Runtime/CMakeFiles/RuntimeTests.dir/AccessTest.cpp.o: in function `createTemporaryFile[abi:cxx11](char const*, (anonymous namespace)::AccessType const&)':
AccessTest.cpp:(.text._ZL19createTemporaryFileB5cxx11PKcRKN12_GLOBAL__N_110AccessTypeE+0x194): undefined reference to `llvm::Twine::str[abi:cxx11]() const'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```
indicates an ABI issue. Maybe using the wrong headers or linking the wrong library? The issue should be with `LLVM_ENABLE_LIBCXX=ON` and probably would affect all LLVM_ENABLE_RUNTIMES.

Meinersbur:
> What could be happening is that the host compiler builds the llvm libraries with libcxx, then the just built clang does not. So the flang-rt test is then trying to combine libstdc++ and libc++.

If that is the case, what should the correct solution be? Tell clang to also use libc++ for the runtimes build?

Meinersbur:
Possible fix: https://github.com/llvm/llvm-project/pull/134990

PR #388, opened 2025-02-22
LLVM_ENABLE_RUNTIMES=flang-rt for flang-aarch64-out-of-tree

Meinersbur:
Make `FlangBuilder.getFlangOutOfTreeBuildFactory` build the flang-rt runtime as an out-of-tree runtimes build with additional steps.

Fixes some issues with FlangBuilder too:
 * Clears the build directories with the clean_obj flag set. Usually buildbot deletes the build directory whenever a CMakeLists.txt in a depends_on_projects subdirectory changes to get a clean build. Otherwise, an incompatible CMakeCache.txt will carry on forever.
 * Add `flang` and `flang-rt` as depends_on_projects. Otherwise, changes in these subdirectories do not trigger a build. Even worse, breaking commits will be attributed to the next commit in llvm/clang/mlir/openmp, as happed in e.g. https://lab.llvm.org/buildbot/#/builders/53/builds/11759 (commit causing failure was https://github.com/llvm/llvm-project/commit/fe8b323f598393d5a7cf468865c4f60d39cb0718)

This PR is not strictly necessary for https://github.com/llvm/llvm-project/pull/124126, it just wouldn't build the runtime anymore.

Affected builders:
 * flang-aarch64-out-of-tree

Affected workers: 
 * linaro-flang-aarch64-out-of-tree

Verfied locally on x86_64.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2393,6 +2393,9 @@ all += [
                     flang_extra_configure_args=[
                         "-DFLANG_ENABLE_WERROR=ON",
                         "-DCMAKE_BUILD_TYPE=Release",
+                    ],
+                    flang_rt_extra_configure_args=[
+                        "-DCMAKE_BUILD_TYPE=Release",
                     ])},
 
     {'name' : "flang-aarch64-debug-reverse-iteration",
--- a/zorg/buildbot/builders/FlangBuilder.py
+++ b/zorg/buildbot/builders/FlangBuilder.py
@@ -1,12 +1,15 @@
 from zorg.buildbot.commands.CmakeCommand import CmakeCommand
 from zorg.buildbot.process.factory import LLVMBuildFactory
 from zorg.buildbot.builders.UnifiedTreeBuilder import addNinjaSteps, getCmakeWithNinjaBuildFactory
+from buildbot.plugins import util, steps
+import os
 
 def getFlangOutOfTreeBuildFactory(
            checks = None,
            clean = False,
            llvm_extra_configure_args = None,
            flang_extra_configure_args = None,
+           flang_rt_extra_configure_args = None,
            env = None,
            **kwargs):
 
@@ -14,7 +17,9 @@ def getFlangOutOfTreeBuildFactory(
         env = dict()
 
     f = getCmakeWithNinjaBuildFactory(
-            depends_on_projects=['llvm','clang','mlir','openmp'],
+            depends_on_projects=['llvm','clang','mlir','openmp','flang','flang-rt'],
+            enable_projects=['llvm','clang','mlir'],
+            enable_runtimes=['openmp'],
             obj_dir="build_llvm",
             checks=[],
             clean=clean,
@@ -24,6 +29,12 @@ def getFlangOutOfTreeBuildFactory(
 
     if checks is None:
         checks = ['check-all']
+
+    cleanBuildRequested = (
+        lambda step: step.build.getProperty("clean")
+        or step.build.getProperty("clean_obj")
+        or clean
+    )
 
     # Make a local copy of the flang configure args, as we are going to modify that.
     if flang_extra_configure_args:
@@ -52,6 +63,16 @@ def getFlangOutOfTreeBuildFactory(
             LLVMBuildFactory.pathRelativeTo(clang_dir, flang_obj_dir)),
         ])
 
+    f.addStep(
+        steps.RemoveDirectory(
+            name=f"clean-{flang_obj_dir}-dir",
+            dir=flang_obj_dir,
+            haltOnFailure=False,
+            flunkOnFailure=False,
+            doStepIf=cleanBuildRequested,
+        )
+    )
+
     # We can't use addCmakeSteps as that would use the path in f.llvm_srcdir.
     f.addStep(CmakeCommand(name="cmake-configure-flang",
                            haltOnFailure=True,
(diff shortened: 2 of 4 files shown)


luporl (commented):
I have tested the changes on AArch64, but manually only, using cmake invocations extracted from these sources. There were no issues in the configure, build and check stages.

Overall the changes look good to me, but as I'm not very familiar with zorg, it would be better if someone else could review this PR too, especially the changes to clean build directories and those around `enable_projects`.

zorg/buildbot/builders/FlangBuilder.py
```diff
@@ -1,20 +1,25 @@
...
            **kwargs):
 
     if env is None:
         env = dict()
 
     f = getCmakeWithNinjaBuildFactory(
-            depends_on_projects=['llvm','clang','mlir','openmp'],
+            depends_on_projects=['llvm','clang','mlir','openmp','flang','flang-rt'],
+            enable_projects=['llvm','clang','mlir'],
+            enable_runtimes=['openmp'],
```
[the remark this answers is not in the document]
Meinersbur: Not every project, build all projects that could cause a build failure of flang/flang-rt. This excludes bolt, lld, lldb, libc++, .... This build mode is used for all buildbot builders. Otherwise the wrong people would get a blame email if a build breaks, not would the actually breaking commit appear in the "Changes" list of the buildbot master UI.

zorg/buildbot/builders/FlangBuilder.py
```diff
@@ -71,4 +92,58 @@ def getFlangOutOfTreeBuildFactory(
...
+        "-DLLVM_ENABLE_RUNTIMES=flang-rt",
+    ]
+
+    # Use the Fortran compiler from the previous step.
+    flang_rt_cmake_args += [
+        util.Interpolate(
+            f"-DCMAKE_Fortran_COMPILER=%(prop:builddir)s/{flang_obj_dir}/bin/flang"
+        ),
+        "-DCMAKE_Fortran_COMPILER_WORKS=ON",
+    ]
```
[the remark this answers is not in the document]
Meinersbur: This would be possible, but not necessary. We already many builder that are using the bootstrapping build (which by design use just-built Clang to build the runtime), keeping at least one that also tests compilation with gcc might be a good idea. The other one `flang-runtime-cuda-gcc` is a bootstrapping build, therefore effectively changed to use Clang to build flang-rt in #393.

buildbot/osuosl/master/config/builders.py
```diff
@@ -2393,6 +2393,9 @@
                     flang_extra_configure_args=[
                         "-DFLANG_ENABLE_WERROR=ON",
                         "-DCMAKE_BUILD_TYPE=Release",
+                    ],
+                    flang_rt_extra_configure_args=[
+                        "-DCMAKE_BUILD_TYPE=Release",
```
DavidSpickett, line 2398: The build type needs to be restated for flang_rt because it's being built in a separate step, do I understand correctly?

zorg/buildbot/builders/FlangBuilder.py
```diff
@@ -71,4 +92,58 @@ def getFlangOutOfTreeBuildFactory(
...
+            env=env,
+            workdir=flang_rt_obj_dir,
+            **kwargs,
+        )
+    )
+
+    addNinjaSteps(
+       f,
+       obj_dir=flang_rt_obj_dir,
+       checks=['check-flang-rt'],
```
DavidSpickett, line 144: I would prefer to have build and test as separate steps. A single step is ok but it's always a bit of extra thinking when a build problem makes a check step red.

zorg/buildbot/process/factory.py
```diff
@@ -73,13 +73,25 @@ def __init__(self, steps=None, depends_on_projects=None, hint=None, **kwargs):
...
         # Build the list of projects to enable.
-        self.enable_projects = \
-            self.depends_on_projects.difference(self.enable_runtimes)
+        enable_projects = kwargs.pop('enable_projects', None)
+        if enable_projects is None or enable_projects == "auto":
+            # Assume that all non-runtimes depends_on_projects dirs are projects.
+            self.enable_projects = \
+                self.depends_on_projects.difference(self.enable_runtimes)
+        elif enable_projects == "all":
+            raise Exception("enable_projects='all' not yet supported")
```
DavidSpickett, line 83: Did you check whether any bot attempts to use all? A quick search in the repo should be enough.

Raising an error is ok if it's unusual but we don't want to break an existing bot.

buildbot/osuosl/master/config/builders.py
Meinersbur: Correct, just like `flang_extra_configure_args`. I could have made getFlangOutOfTreeBuildFactory do it by default, but it didn't to it for the flang step either.

zorg/buildbot/builders/FlangBuilder.py
Meinersbur: `addNinjaSteps` is adding multiple steps, indicated by its plural name. Or see its code: https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/builders/UnifiedTreeBuilder.py#L215-L242.

zorg/buildbot/process/factory.py
Meinersbur: Since I added that function argument, there cannot be any existing callers that use that argument. If there was, the buildbot would fail to startup giving that exception as error message, which I did not observe.

zorg/buildbot/builders/FlangBuilder.py
DavidSpickett: I see, so this will `ninja` then `ninja check-flang-rt`. That's fine.

DavidSpickett (commented):
LGTM

(and thanks for all the effort you've been putting into this!)

DavidSpickett (approved):

PR #392, opened 2025-02-24
Revert "[GreenDragon] Workaround Stage 2 bots being broken (#163)"

cyndyishida:
This reverts commit 3ed3d915a9d4de7967dd20a35a9a266b09cdba86.

Now that the CI machines have newer Xcode's and more importantly SDKs that contain stable version of libc++, re-enable module build so we can catch breakages from upstream faster.

resolves: rdar://125128256


cyndyishida:
@justice-adams-apple  It looks like you are not apart of the llvm-org but can you please review too?

justice-adams-apple (approved):

PR #398, opened 2025-03-03
Add flang-arm64-windows-msvc buildbot

omjavaid:
This patch introduces a dedicated flang-arm64-windows-msvc buildbot to decouple Flang from the existing Clang buildbot splitting load.

This will test flang with openmp and flang-rt runtimes and will run check-mlir, check-flang, and check-flang-rt.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2541,6 +2541,21 @@ all += [
                         '-DLLVM_PARALLEL_COMPILE_JOBS=4',
                     ])},
 
+    {'name': "flang-arm64-windows-msvc",
+    'tags' : ["mlir", "flang"],
+    'workernames' : ["linaro-armv8-windows-msvc-01"],
+    'builddir': "flang-arm64-windows-msvc",
+    'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
+                    depends_on_projects=['llvm', 'clang', 'lld', 'mlir', 'compiler-rt', 'openmp', 'flang','flang-rt'],
+                    checks=['check-mlir', 'check-flang', 'check-flang-rt'],
+                    extra_configure_args=[
+                        "-DLLVM_TARGETS_TO_BUILD=X86;AArch64",
+                        "-DCLANG_DEFAULT_LINKER=lld",
+                        "-DCMAKE_TRY_COMPILE_CONFIGURATION=Release",
+                        "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
+                        "-DCMAKE_C_COMPILER_LAUNCHER=ccache",
+                        "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache"])},
+
     {'name' : 'ppc64-flang-aix',
     'tags'  : ["flang", "ppc", "ppc64", "aix"],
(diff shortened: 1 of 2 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -2541,6 +2541,21 @@
                         '-DLLVM_PARALLEL_COMPILE_JOBS=4',
                     ])},
 
+    {'name': "flang-arm64-windows-msvc",
+    'tags' : ["mlir", "flang"],
+    'workernames' : ["linaro-armv8-windows-msvc-01"],
+    'builddir': "flang-arm64-windows-msvc",
+    'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
+                    depends_on_projects=['llvm', 'clang', 'lld', 'mlir', 'compiler-rt', 'openmp', 'flang','flang-rt'],
```
DavidSpickett, line 2549: Do the runtimes names here end up passed to LLVM_ENABLE_RUNTIMES?

DavidSpickett (commented):
You said this is to spread the work. I don't recall out of memory errors on Windows on Arm, which is the usual problem we have with Flang, so is it purely so that Clang can build more frequently?

You could also argue that we have separate Clang and Flang bots for Linux, so this change aligns Windows with that.

Also this is single stage, right? So the current 2 stage bot will still be building Flang with stage 1's Clang. That seems valauble enough to keep Flang enabled there.

omjavaid:
> You said this is to spread the work. I don't recall out of memory errors on Windows on Arm, which is the usual problem we have with Flang, so is it purely so that Clang can build more frequently?

This is not to divide the load based on memory usage but rather done to split the testsuite execution time. The common observation in recent months is that more often we have a failing flang test or build failure in flang rather than rest of the llvm project but check llvm and check clang take most of the time during testsuite check-all. With this builder we will be able to run flang testsuite early hence catching problems way earlier.
 
> You could also argue that we have separate Clang and Flang bots for Linux, so this change aligns Windows with that.

Agreed.
 
> Also this is single stage, right? So the current 2 stage bot will still be building Flang with stage 1's Clang. That seems valauble enough to keep Flang enabled there.

Yes for now there is no plan to change the 2 stage bot although it has a long execution time but that can only be fixed with better hardware.

DavidSpickett:
I didn't even consider the testing time. With all the given reasons, I agree this is a good change.

DavidSpickett (approved):
LGTM

buildbot/osuosl/master/config/builders.py
```diff
@@ -2541,6 +2541,21 @@
...
+                    checks=['check-mlir', 'check-flang', 'check-flang-rt'],
+                    extra_configure_args=[
```
luporl, line 2551: It would be nice to enable assertions, to catch more issues: `-DLLVM_ENABLE_ASSERTIONS=ON`

luporl (approved):
LGTM

ceseo (approved):

buildbot/osuosl/master/config/builders.py
omjavaid: UnifiedTreeBuilder enables DLLVM_ENABLE_ASSERTIONS=ON by default so we dont have to mention it explicitly in the configure flags.

PR #402, opened 2025-03-04
LLVM_ENABLE_RUNTIMES=flang-rt for amdgpu-offload-*

Meinersbur:
Add `depends_on_projects=['flang-rt']`, and `checks=['check-flang-rt']` to the ppc64le-flang-rhel-clang builder. The prepares the removal of the "projects" build of the flang runtime in https://github.com/llvm/llvm-project/pull/124126.

The corresponding change in the LLVM repository is https://github.com/llvm/llvm-project/pull/129692

Affected builders (production):
 * [amdgpu-offload-ubuntu-22-cmake-build-only](https://lab.llvm.org/buildbot/#/builders/203)
 * [amdgpu-offload-rhel-9-cmake-build-only](https://lab.llvm.org/buildbot/#/builders/205)
 * [amdgpu-offload-rhel-8-cmake-build-only](https://lab.llvm.org/buildbot/#/builders/204)

Affected workers (production):
 * [rocm-docker-ubu-22](https://lab.llvm.org/buildbot/#/workers/162)
 * [rocm-docker-rhel-9](https://lab.llvm.org/buildbot/#/workers/163)
 * [rocm-docker-rhel-8](https://lab.llvm.org/buildbot/#/workers/164)

Admins listed for those workers:
 * AMD <dl.gcr.lightning.buildbot@amd.com>

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1961,7 +1961,7 @@ all += [
     'builddir': "amdgpu-offload-ubuntu-22-cmake-build-only",
     'collapseRequests' : False,
     'factory' : AnnotatedBuilder.getAnnotatedBuildFactory(
-                    depends_on_projects=["llvm", "clang", "flang", "mlir", "lld", "compiler-rt", "libcxx", "libcxxabi", "openmp", "offload", "libunwind"],
+                    depends_on_projects=["llvm", "clang", "flang", "flang-rt", "mlir", "lld", "compiler-rt", "libcxx", "libcxxabi", "openmp", "offload", "libunwind"],
                     script="amdgpu-offload-cmake.py",
                     checkout_llvm_sources=True,
                     script_interpreter=None
@@ -1973,7 +1973,7 @@ all += [
     'builddir': "amdgpu-offload-rhel-9-cmake-build-only",
     'collapseRequests' : False,
     'factory' : AnnotatedBuilder.getAnnotatedBuildFactory(
-                    depends_on_projects=["llvm", "clang", "flang", "mlir", "lld", "compiler-rt", "libcxx", "libcxxabi", "openmp", "offload", "libunwind"],
+                    depends_on_projects=["llvm", "clang", "flang", "flang-rt", "mlir", "lld", "compiler-rt", "libcxx", "libcxxabi", "openmp", "offload", "libunwind"],
                     script="amdgpu-offload-cmake.py",
                     checkout_llvm_sources=True,
                     script_interpreter=None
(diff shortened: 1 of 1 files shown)


Meinersbur:
Unfortunately, the annotated builders are impossible to check locally using instructions from https://llvm.org/docs/HowToAddABuilder.html#testing-a-builder-config-locally, as the build instructions are taken directly from multiple main directories on github.com:
 * The buildbot master takes the builder [configuration](https://github.com/llvm/llvm-zorg/blob/872f477610d83821c9f1368c969006789b21011b/buildbot/osuosl/master/config/builders.py#L1965) which contain the instruction for the worker to execute `amdgpu-offload-cmake.py`.
 * The worker checks out [llvm-zorg](https://github.com/llvm/llvm-zorg/tree/main) from its public repository and executes [amdgpu-offload-cmake.py](https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py) from there (NOT the locally modified version of it). It contains [instructions](https://github.com/llvm/llvm-zorg/blob/872f477610d83821c9f1368c969006789b21011b/zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py#L26) to pre-populate the CMake cache from file `AMDGPUBot.cmake`.
 * The worker also checks out [llvm-project](https://github.com/llvm/llvm-project/tree/main) from its public repository which contains the [AMDGPUBot.cmake](https://github.com/llvm/llvm-project/blob/main/offload/cmake/caches/AMDGPUBot.cmake) file.

Running the unchanged/`main` worktree gets me the following error:
```
Running: cmake ../llvm-project/llvm -GNinja "-C ../llvm-project/offload/cmake/caches/AMDGPUBot.cmake"
Current working directory cannot be established.
['cmake', '../llvm-project/llvm', '-GNinja', '-C ../llvm-project/offload/cmake/caches/AMDGPUBot.cmake'] exited with return code 1.
```
I think this is because `amdgpu-offload-cmake.py` is [deleting its own current work directory](https://github.com/llvm/llvm-zorg/blob/872f477610d83821c9f1368c969006789b21011b/zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py#L22) [sic!], after which the cwd file descriptor refers to a non-existing directory which CMake cannot cope with. I cannot verify this assumption without modifying the upstream repositories.

jplehr:
I have a setup for local testing and can take a swing at it for a test.

jplehr:
> I think this is because `amdgpu-offload-cmake.py` is [deleting its own current work directory](https://github.com/llvm/llvm-zorg/blob/872f477610d83821c9f1368c969006789b21011b/zorg/buildbot/builders/annotated/amdgpu-offload-cmake.py#L22) [sic!], after which the cwd file descriptor refers to a non-existing directory which CMake cannot cope with.

That is not quite the intended behavior, but after looking at the code, I believe you are right. Let me put up a patch that does what it should do more reasonably.

PR #403, opened 2025-03-04
[AMDGPU] Change CWD from within tmp directory

jplehr:
As pointed out in https://github.com/llvm/llvm-zorg/pull/402 the former solution deleted the CWD from within that directory, rendering the state quite useless and resulting in errors on execution.

The approach in here creates a temp directory, changes to it, removes the old CWD, recreates it, and changes back to that directory. I opted for this solution as, oddly enough, I had some trouble removing the
  enumerated list of files in the buildtree.


jhuber6 (approved):

Meinersbur:
One could also have deleted all the contents of the directory, instead the directory itself.

PR #405, opened 2025-03-04
[HIP] Builder for third-party libraries (Kokkos first)

jplehr:
This adds:
- Builder for HIP-enabled Kokkos library build+test
- Worker for that builder
- Script for AnnotatedBuilder to build third-party libraries

This increases HIP test coverage substantially in the upstream and helps to identify issues earlier.
Initially, this will contain only Kokkos.
The Ginkgo library is lined up and will be added next.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1991,6 +1991,18 @@ all += [
                     script_interpreter=None
                 )},
 
+    # This one has a longer turn-around time, so we cannot disallow collapsing requests
+    {'name' : "hip-third-party-libs-test",
+    'tags'  : ["amdgpu", "offload", "openmp"],
+    'workernames' : ["ext_buildbot_hw_05-hip-docker"],
+    'builddir': "hip-third-party-libs-test",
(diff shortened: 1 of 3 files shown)


yxsamliu (approved):

buildbot/osuosl/master/config/builders.py
```diff
@@ -1991,6 +1991,19 @@
                     script_interpreter=None
                 )},
 
+    # This one has a longer turn-around time, so we cannot disallow collapsing requests
+    {'name' : "hip-third-party-libs-test",
+    'tags'  : ["amdgpu", "offload", "openmp"],
+    'workernames' : ["ext_buildbot_hw_05-hip-docker"],
+    'builddir': "hip-third-party-libs-test",
+    'collapseRequests' : True,
```
gkistanova, line 1999: This would make the builder collapse all the build requests unconditionally. You do not want this, Just remove this argument and let the default machinery do the work.

gkistanova (changes requested):
Thanks for the patch, JP.

It looks fine with one small thing to address.

PR #407, opened 2025-03-12
[openmp-offload-amdgpu-clang-flang] Re-add FLANG_RUNTIME_F128_MATH_LIB

Meinersbur:
The option was removed in #395. After https://github.com/llvm/llvm-project/pull/130411, the option should now work even when compiling flang-rt with Clang.


jplehr (approved):
LGTM

This builder is in staging, so we can land and revert if this change breaks the build.

PR #409, opened 2025-03-21
[AMDGPU] Adjust xfail tests

jplehr:
The containerized runner is executing on a certain CPU set. This seems to confuse the specific test and makes it fail.
xfail it therefore.


jplehr:
Hi @gkistanova getting this in is somewhat urgent/important as we need to switch one of our builders over. Can we get this into the production runner maybe even today?
Thank you a lot and apologies for the short notice.

jhuber6 (approved):

PR #411, opened 2025-03-26
Update lldb jobs to run on arm 64 machines

justice-adams-apple:
Update lldb-matrix and lldb-sanitized to run on apple SI machines

### DIFF
--- a/zorg/jenkins/jobs/jobs/lldb-cmake-matrix
+++ b/zorg/jenkins/jobs/jobs/lldb-cmake-matrix
@@ -102,7 +102,7 @@ pipeline {
                       --runtimes="libcxx;libcxxabi;libunwind"  \
                       --cmake-type=Release \
                       --dotest-flag="--skip-category gmodules" \
-                      --dotest-flag="--arch=aarch64"
+                      --dotest-flag="--arch=arm64"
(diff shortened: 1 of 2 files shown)


zorg/jenkins/jobs/jobs/lldb-cmake-matrix
```diff
@@ -21,7 +21,7 @@ pipeline {
         }
     }
     environment {
-        MATRIX_COMMON_BUILD_PARAMETERS = '--cmake-flag=-DLLVM_TARGETS_TO_BUILD=X86'
+        MATRIX_COMMON_BUILD_PARAMETERS = '--cmake-flag=-DLLVM_TARGETS_TO_BUILD=AArch64'
```
Michael137, line 24: Hmm I think we want this to be `X86;AArch64`

Only specifying one of them means we wouldn't run some of the x86 tests on ARM machine and vice-versa. But we do want that test-coverage. They don't actually require an x86 host

Michael137 (approved):
🚢
Thanks!

Just minor comment re. CMake change

adrian-prantl:
LGTM with the discussed changes. Let's try it out!

PR #412, opened 2025-03-26
Pass -no_fixup_chains to the linker

ahatanak:
This is needed to work around a limitation of the linker, which was causing it to crash on coverage-enabled bots.

rdar://146130000


azharudd (approved):
We can try this. Thanks!

PR #414, opened 2025-03-27
Update setup of LNT to handle python 3.12+

maxim-kuvyrkov:
Virtualenv in python 3.12+ no longer installs setuptools by default, see [1].  This patch adds a "pip install setuptools" step to ClangBuilder.py.

Note that running
$ VIRTUALENV_SETUPTOOLS=bundle virtualenv sandbox
will install an older/incompatible version of setuptools, at least on Ubuntu 24.04 system.

[1] https://virtualenv.pypa.io/en/latest/user_guide.html#seeders

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -615,6 +615,12 @@ def _getClangCMakeBuildFactory(
                                description='recreating sandbox',
                                workdir='test',
                                env=env))
+        f.addStep(ShellCommand(name='install lnt dependencies',
+                               command=[python, '-m', 'pip', 'install', 'setuptools'],
+                               haltOnFailure=True,
+                               description='install lnt dependencies',
+                               workdir='test/sandbox',
+                               env=env))
         f.addStep(ShellCommand(name='setup lit',
                                command=[python, lnt_setup, 'develop'],
                                haltOnFailure=True,


zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -615,6 +615,12 @@ def _getClangCMakeBuildFactory(
                                description='recreating sandbox',
                                workdir='test',
                                env=env))
+        f.addStep(ShellCommand(name='install lnt dependencies',
+                               command=[python, '-m', 'pip', 'install', 'setuptools'],
```
DavidSpickett, line 619: Given that "setup lit" uses the sandbox's Python:
```
/home/tcwg-buildbot/worker/clang-aarch64-sve-vla/test/sandbox/bin/python /home/tcwg-buildbot/worker/clang-aarch64-sve-vla/test/lnt/setup.py develop
```
I think we are installing setuptools into the sandbox itself here.

Which is what we want to do, so we don't conflict with existing installs.

It's just not very clear from `workdir=` whether it will activate the venv, or magically make that the PATH. I guess it puts that on the PATH somehow.

Anyway, whatever it does, it works.

DavidSpickett (commented):
Is ClangBuilder the only one setting up LNT, or are you purposely limiting the scope here?

If it's the latter, note in the title that this is only for ClangBuilder.

zorg/buildbot/builders/ClangBuilder.py
maxim-kuvyrkov: The env is activated by calling python from $env/bin directory.  This is arrange through "python", which is a variable, not a string.

maxim-kuvyrkov:
> Is ClangBuilder the only one setting up LNT, or are you purposely limiting the scope here?
> 
> If it's the latter, note in the title that this is only for ClangBuilder.

I found references to "virtualenv" only in ClangBuilder.  I'll update the commit message.  Thanks.

maxim-kuvyrkov:
Resolved @DavidSpickett 's review comments and updated PR description.

DavidSpickett (approved):

DavidSpickett:
I thought maybe this would break older virtualenv, but pip should just find that setuptools is already installed, so no problem there.

DavidSpickett:
I have the ability to merge here, I assume you do not. Ok to merge?

PR #416, opened 2025-03-27
AIX: Default SANITIZER_DISABLE_SYMBOLIZER_PATH_SEARCH to ON

hubert-reinterpretcast:
llvm/llvm-project#129012 added a new CMake flag. This flag is used by IBM's downstream builds.

This patch sets the flag on the community buildbot for AIX to ensure upstream coverage and visibility.


daltenty:
I suggest we do flang as well.

hubert-reinterpretcast:
> I suggest we do flang as well.

Done.

daltenty (approved):
LGTM

hubert-reinterpretcast:
@gkistanova, fyi re: newly merged config change (not urgent) that will get picked up by the next restart.

PR #418, opened 2025-03-31
[presubmit] update arc version

Keenuts:
Updating the running version of GHA controller/listener/runner to the latest.

### DIFF
--- a/premerge/main.tf
+++ b/premerge/main.tf
@@ -216,7 +216,7 @@ resource "helm_release" "github_actions_runner_controller" {
   name       = "llvm-premerge-controller"
   namespace  = "llvm-premerge-controller"
   repository = "oci://ghcr.io/actions/actions-runner-controller-charts"
-  version    = "0.9.3"
+  version    = "0.11.0"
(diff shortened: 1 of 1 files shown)


lnihlen (approved):

boomanaiden154:
I can land this Friday evening when traffic is low if that is okay with everyone else.

Keenuts:
> I can land this Friday evening when traffic is low if that is okay with everyone else.

the cluster was already updated to 0.11.0 while fixing yesterday's mess, so we can merge this to match the cluster state now no?

boomanaiden154:
I ended up reverting that change as it doesn’t fully go through unless you uninstall and reinstall the helm charts, so the cluster is currently at the state of what is in main.

Keenuts:
> I ended up reverting that change as it doesn’t fully go through unless you uninstall and reinstall the helm charts, so the cluster is currently at the state of what is in main.

Ah good to know, in that case yes let's wait for friday

boomanaiden154 (approved):

PR #419, opened 2025-03-31
[libc][AMDGPU] Add libc to the runtimes list

jplehr:
It appears that if something is contained in the `depends_on_project` list but is not explicitly mentioned as an `enable_runtimes`, while also an explicit `enable_runtimes` list is passed to the unified tree builder, that component is auto-added as an `LLVM_ENABLE_PROJECTS`. This is wrong in the case of libc.

Wile we only care about the version being built for AMDGPU, this appears to be the easiest fix to unblock our libc bot.


jplehr:
@gkistanova  I believe the bot in question is in production. I'd appreciate if we can propagate this to prod, once landed.

jhuber6 (approved):

gkistanova:
This change has been applied to production.

PR #420, opened 2025-03-31
Add documentation on resetting ARC

boomanaiden154:
Upgrading/resetting ARC is a somewhat involved process that cannot be done with
just terraform apply with an updated configuration. This patch adds some
documentation on the exact steps to perform, some of the motivation behind
them, and some tips and tricks that I have developed while hacking on the
cluster.

### DIFF
--- a/premerge/cluster-management.md
+++ b/premerge/cluster-management.md
@@ -98,3 +98,107 @@ Setting the cluster up for the first time is more involved as there are certain
 resources where terraform is unable to handle explicit dependencies. This means
 that we have to set up the GKE cluster before we setup any of the Kubernetes
 resources as otherwise the Terraform Kubernetes provider will error out.
+
+## Upgrading/Resetting Github ARC
+
+Updating and resetting the Github Actions Runner Controller (ARC) within the
+cluster involves largely the same process. Some special considerations need
+to be made with how ARC interacts with kubernetes. The process involves
+uninstalling the runner scale set charts, deleting the namespaces to ensure
+everything is properly cleaned up, optionally bumping the version number if
+this is a version upgrade, and then reinstalling the charts to get the cluster
+back to accepting production jobs.
+
+It is important to not just blindly delete controller pods or namespaces as
+this (at least empirically) can interrupt the state and custom resources that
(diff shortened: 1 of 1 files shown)


premerge/cluster-management.md
```diff
@@ -98,3 +98,107 @@ Setting the cluster up for the first time is more involved as there are certain
...
+uninstalling the runner scale set charts, deleting the namespaces to ensure
+everything is properly cleaned up, optionally bumping the version number if
+this is a version upgrade, and then reinstalling the charts to get the cluster
+back to accepting production jobs.
+
+It is important to not just blindly delete controller pods or namespaces as
+this (at least empirically) can interrupt the state and custom resources that
+ARC manages, then requiring a costly full uninstallation and reinstallation of
+at least a runner scale set.
+
```
lnihlen, line 116: Can we get a bit here about impact on any running build jobs?

premerge/cluster-management.md
```diff
@@ -98,3 +98,107 @@ Setting the cluster up for the first time is more involved as there are certain
...
+This is not necessary only for bumping the version of ARC. This involves simply
+updating the version field for the `helm_release` objects in `main.tf`. Make sure
+to commit the changes and push them to `llvm-zorg` to ensure others working on
+the terraform configuration have an up to date state when they pull the repository.
+
+### Bringing the Cluster Back Up
+
+To get the cluster back up and accepting production jobs again, simply run
+`terraform apply`. It will recreate all the resource previously destroy and
```
lnihlen, line 198: s/destroy/destroyed

premerge/cluster-management.md
boomanaiden154: Best practice is to just uninstall the helm charts. The jobs currently running will still finish up, and ARC will not accept any more jobs. If you set `maxRunners` to zero, ARC might accept some jobs but never start executing them. I don't think that would cause any problems in our configuration, but for HA configurations with multiple clusters, it can cause non-seamless failover from what I understand.

https://www.kenmuse.com/blog/strategies-for-upgrading-arc/

boomanaiden154: Added some comments on this.

PR #421, opened 2025-03-31
Add documentation on 3-31-25 incident

boomanaiden154:
This patch adds at least some initial documentation on the incident that
happened this morning, primarily linking to the more detailed instructions
in the cluster management documentation on how it was resolved by
uninstalling and reinstalling the Helm charts.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1953,7 +1953,7 @@ all += [
     'builddir': "openmp-offload-amdgpu-runtime-2",
     'factory' : OpenMPBuilder.getOpenMPCMakeBuildFactory(
                         clean=True,
-                        enable_runtimes=['compiler-rt', 'libunwind', 'libcxx', 'libcxxabi', 'openmp', 'offload'],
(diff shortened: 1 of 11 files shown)


lnihlen (approved):

lnihlen:
Did we lose any queued jobs? And do we have any long-term plans to address this issue so it doesn't recur?

Keenuts:
> Did we lose any queued jobs? And do we have any long-term plans to address this issue so it doesn't recur?

I manually cancelled 2 jobs during one of the initial cleanups, but otherwise jobs were just queued for a long time and not lost.

gburgessiv (approved):
Thanks for writing this (& especially the supporting docs in your other PR)!

The other PR is next on my review-list; maybe you address this there, but is there any known follow-up action we can take here to reduce(/eliminate) the chances of this? Or is this something like "GH or Cloud had a general unrecoverable issue that we had to turn everything off and on again to get around"?

boomanaiden154:
> but is there any known follow-up action we can take here to reduce(/eliminate) the chances of this

Something we're still investigating, but I would presume yes.

PR #424, opened 2025-04-02
Reduce system resource usage for flang-runtime-cuda-* builders on as-builder-7 worker.

vvereschaka:
The latest changes for 'flang' and 'flang-rt' projects cause the build host overload and freeze the system for a long time. Reduce resource usage for the build to avoid it.

Note: restore the changes after the problem will get solved.


gkistanova (approved):
LGTM.

vvereschaka:
I repalced `UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory` call with `UnifiedTreeBuilder.getCmakeExBuildFactory`. The first one does not support the `jobs` argument.

gkistanova (approved):
The first one supports `jobs` as a build property. But this one seems fine too.

PR #429, opened 2025-04-14
[lldb] Print cmake version in node info

justice-adams-apple:
Print cmake version as part of node info stage


justice-adams-apple:
cc @Michael137

Michael137 (approved):

PR #430, opened 2025-04-16
[fuchsia] Downgrade cmake from latest to release from Ubuntu jammy

zeroomega:
The CMake 4.0 no longer allow adding a shared library target when
the target platform is a baremetal platform. This change downgrades
the cmake from 4.0 to jammy release (3.22) to mitigate this issue.


petrhosek (approved):

zeroomega:
I don't have a local docker setup to test the build. But it should work.
The cmake from ubuntu jammy is https://launchpad.net/ubuntu/jammy/+source/cmake , which has version number 3.22.

PR #432, opened 2025-04-17
Fix flang-aarch64-libcxx builder

luporl:
After flang-rt, flang-aarch64-libcxx builder started to fail.
Before it, llvm libraries, flang and its runtime were built with the
host compiler, but now flang runtime is built with the stage 1 clang.
The problem is that the C++ library of these compilers may be
incompatible. The linked issue has more details.

To avoid this issue, build flang-rt with the host compiler.

Fixes https://github.com/llvm/llvm-project/issues/135381

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2489,6 +2489,9 @@ all += [
                         "-DLLVM_ENABLE_ASSERTIONS=ON",
                         "-DLLVM_ENABLE_LIBCXX=On",
                         "-DCMAKE_BUILD_TYPE=Release",
+                        "-DLLVM_RUNTIME_TARGETS=aarch64-unknown-linux-gnu",
+                        "-DRUNTIMES_aarch64-unknown-linux-gnu_CMAKE_CXX_COMPILER='c++'",
+                        "-DRUNTIMES_aarch64-unknown-linux-gnu_CMAKE_C_COMPILER='cc'"
                         ])},
 
     {'name' : "flang-aarch64-release",
--- a/zorg/buildbot/builders/FlangBuilder.py
+++ b/zorg/buildbot/builders/FlangBuilder.py
@@ -138,10 +138,22 @@ def getFlangOutOfTreeBuildFactory(
         )
     )
 
+    # 'check-flang-rt' becomes 'check-flang-rt-<target>' when a runtime target
+    # is specified.
+    check_flang_rt = 'check-flang-rt'
+    for arg in llvm_extra_configure_args:
+        if arg.find("DLLVM_RUNTIME_TARGETS") != -1:
+            targets = arg.split('=', 1)[1]
+            # Currently only one target is supported.
+            if not targets or targets.find(";") != -1:
+                break
+            check_flang_rt += '-' + targets
+            break
+
     addNinjaSteps(
        f,
        obj_dir=flang_rt_obj_dir,
-       checks=['check-flang-rt'],
+       checks=[check_flang_rt],
        env=env,
        stage_name="flang-rt",
        **kwargs)


luporl:
Marked as draft because the changes were only tested in isolation:
- Built and tested flang/flang-rt with the modified flags.
- Called `getFlangOutOfTreeBuildFactory` from a test program, passing flang-aarch64-libcxx flags as `llvm_extra_configure_args`.

Meinersbur:
I am not sure this can work and if it does it will be fragile. The reason is that the LLVM_ENABLE_RUNTIMES system itself sets [CMAKE_CXX_COMPILER](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/LLVMExternalProjectUtils.cmake#L227-L235) and which definition wins is nothing I would rely on.

If a bootstrapping build is not possible, consider a standalone-runtimes build, like flang-runtime-cuda-clang does.

luporl:
> This current PR doesn't seem to use getFlangOutOfTreeBuildFactory for flang-aarch64-libcxx, but might be a solution as well.

Right, flang-aarch64-libcxx uses `getCmakeWithNinjaBuildFactory` instead, I completely missed it...

DavidSpickett:
Thanks for taking this up @luporl .

> Why not adding -I<path-to-libcxx> to -DRUNTIMES_CMAKE_ARGS=-DCMAKE_CXX_FLAGS= so stage1 clang is able to find libcxx?

This works but requires us to 1. hardcode paths or 2. add stable symlinks to our docker container. Both are possible but not great for anyone external who needs to reproduce.

A standalone runtimes build would be ideal, and most "in the spirit" of the runtimes build. Assuming we can find a neat way to do it.

Using the out of tree builder means that flang is also out of tree, I think, so this bot becomes out of tree + libcxx. Normally I wouldn't want to mix concerns but we do have an out of tree bot to compare to at least. It would be clear if the problem was due to using libcxx.

Maybe you can "out of tree" just flang-rt.

buildbot/osuosl/master/config/builders.py
```diff
@@ -2526,21 +2526,30 @@
     'tags'  : ['flang'],
     'workernames' : ["linaro-flang-aarch64-libcxx"],
     'builddir': "flang-aarch64-libcxx",
-    'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
-                    clean=True,
-                    checks=['check-flang','check-flang-rt'],
-                    depends_on_projects=['llvm','mlir','clang','flang','flang-rt'],
-                    extra_configure_args=[
+    'factory' : FlangBuilder.getFlangOutOfTreeBuildFactory(
+                    checks=['check-flang'],
```
DavidSpickett, line 2530: What happened to check-flang-rt, is it added implicitly?

buildbot/osuosl/master/config/builders.py
```diff
@@ -2526,21 +2526,30 @@
...
+                    llvm_extra_configure_args=[
                         "-DLLVM_TARGETS_TO_BUILD=AArch64",
-                        "-DLLVM_INSTALL_UTILS=ON",
                         "-DCMAKE_CXX_STANDARD=17",
                         "-DLLVM_ENABLE_WERROR=OFF",
+                        "-DBUILD_SHARED_LIBS=ON",
```
DavidSpickett: Are we enabling shared libs because the previous configuration did so, or do we actually need it?

I could understand needing it for llvm so that the subsequent flang build can use them, but further into the build it does not seem necessary.

(I don't care about how long it takes, just about being as close as we can to the defaults)

buildbot/osuosl/master/config/builders.py
luporl: Yes, it is added implicitly.
As `flang-aarch64-out-of-tree` builder has only `check-flang`, I used the same for `flang-aarch64-libcxx`.

buildbot/osuosl/master/config/builders.py
luporl: Shared libs are being enabled just because the previous configuration did so. It's probably not needed.

If no errors occur, do you think it would be better to avoid shared libs in all builds (llvm, flang, flang-rt)?

DavidSpickett: Maybe we did this as some way to save build time? I'm not sure if it actually does though.

Yes let's avoid them here unless the alternative has errors or lots more configuration changes. (any other bot can stay as it is)

luporl: There were no errors with no shared libs.

It's hard to compare build times with ccache enabled, but my test build without shared libs took about the same time as the last one, with them enabled.

luporl:
Thanks @DavidSpickett and @Meinersbur, for the reviews and suggestions.

DavidSpickett:
Thanks for the fix Leandro, it's now building correctly - https://lab.llvm.org/staging/#/builders/46/builds/4084.

1 failure but we know this is due to https://github.com/llvm/llvm-zorg/pull/482.

PR #439, opened 2025-04-29
Add "--timeout=900" to builders to avoid hanging tests from causing testing to hang.

dyung:
Add "--timeout=900" to our bots to prevent a hanging test from causing the entire testing step from hanging (for example https://lab.llvm.org/buildbot/#/builders/174/builds/16853). It was suggested that I could add this argument to prevent the problem from happening.


gkistanova (approved):
Let's see if this works around the issue.

PR #445, opened 2025-05-09
[CI] Add a second premerge cluster

boomanaiden154:
We have quota in us-west8 that we want to use for an additional premerge
cluster for a HA setup and to improve capacity. This patch makes the infra
run a cluster on the new quota.


premerge/main.tf
```diff
@@ -61,6 +61,14 @@ moved {
   to   = module.premerge_cluster1
 }
 
+module "premerge_cluster2" {
```
lnihlen, line 64: does it make sense to name them for regions instead of numbers? like "premerge_cluster_west?"

lnihlen (approved):

premerge/main.tf
boomanaiden154: That's a better naming scheme than just numbers. I'll update it.

PR #448, opened 2025-05-15
[CI] Add postmortem info on latest issue

boomanaiden154:
We now have much better information on the runner set not scaling issues. Add some documentation to the issues page about this for the benefit of those that come along and read it later.


gburgessiv (approved):
Thanks for writing this up!

PR #449, opened 2025-05-18
[CI] Update docs after renaming of llvm-premerge-prototype

boomanaiden154:
This patch updates the cluster management documentation now that we have renamed the llvm-premerge-prototype cluster to llvm-premerge-cluster-us-central.


gburgessiv (approved):
thanks for this!

PR #450, opened 2025-05-20
[s390x] Make stage 2 builder use clean builds

uweigand:
Use clean builds for the s390x multistage builder as was done for Linaro in https://github.com/llvm/llvm-zorg/pull/306.  This will enable collapsing build requests to help the builder keep up.

As further speed-up, skip the stage 1 testing, as is already done for most other multistage builders.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -805,9 +805,10 @@ all = [
     'builddir': "clang-s390x-linux-multistage",
     'factory' : ClangBuilder.getClangCMakeBuildFactory(
                     jobs=4,
-                    clean=False,
+                    clean=True,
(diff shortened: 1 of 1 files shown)


DavidSpickett (approved):
I see `-DLLVM_CCACHE_BUILD=ON` in the stage 1 configuration, so this shouldn't add much if any overhead, and https://lab.llvm.org/buildbot/#/builders/42 covers the stage 1 testing already.

LGTM.

DavidSpickett:
https://github.com/llvm/llvm-zorg/issues/250 was the origin of this. We never did find out why it was not an issue before then, I suspect it should have been an issue, but for a bug elsewhere possibly in Buildbot itself that got fixed at some point.

PR #451, opened 2025-05-21
Update configuration for llvm-clang-x86_64-expensive-checks-ubuntu* builders

vvereschaka:
* switch to clang-21/lld-21 instead of gcc/ld.gold to speed up the builds and avoid the ld.gold liner warnings related with unsupported DWARF version.
* switch to release build to speed up the builds because of clang/llvm table-gen tool.

Also moved to 'UnifiedTreeBuilder.getCmakeExBuildFactory' factory.


andreil99 (approved):
Thanks, Vlad!

PR #453, opened 2025-05-22
Skip LNT submission

justice-adams-apple:
Skip LNT submission so that the jobs aren't reporting failure status. There is not a stable LNT server setup as of right now

This will allow us to at least get build status from the green dragon LNT jobs


cachemeifyoucan (approved):
LGTM

PR #461, opened 2025-06-06
[CI] Remove Buildkite Token from metrics container

boomanaiden154:
Now that buildkite has been sunsetted, we no longer need to keep track of the Buildkite workflows. https://github.com/llvm/llvm-project/pull/143049 does this on the metrics container side. This patch adjusts the deployment to those changes. Note that patch needs to land first before this can be applied.


cmtice (approved):

boomanaiden154:
This will need to be applied to the cluster after the steps in #143049 are done.

It should be a simple
```
cd premerge
terraform apply
```

(assuming the output of `terraform apply` makes sense).

PR #462, opened 2025-06-06
[BOLT] Add nfc-check-validation step

paschalis-mpeis:
NFC-mode runs in-tree and out-of-tree (large BOLT tests) when there are changes
affecting llvm-bolt between the current revision and the previous one.

However, the revision SHA itself and the GNU build-id may be two unique
differences when doing such a comparison.

This patch introduces a lightweight sanity check step that reports a warning on
the Buildbot so that we are aware when this occurs.


paschalis-mpeis:
Forced push to rebase, as base branch now **renames** the existing Meta buildbots instead of dropping them.
I also slightly reformatting of the output warning message.

paschalis-mpeis:
(forced push to rebase on the parent branch. no other changes here)

PR #465, opened 2025-06-09
[BOLT] Excluding GNU build-id and cleanup.

paschalis-mpeis:
Pass '--build-id' to the linker to disable emission of a unique build ID into a note section. This breaks nfc-mode tests.

Additional cleanup:
- Out-of-tree tests are now labeled as 'BOLT large tests'
- Dropped step 'nfc-stat-check' as it was disabled
- Removed unused flags for Non-NFC tests

### DIFF
--- a/zorg/buildbot/builders/UnifiedTreeBuilder.py
+++ b/zorg/buildbot/builders/UnifiedTreeBuilder.py
@@ -233,17 +233,18 @@ def addNinjaSteps(
     else:
         check_env = env or {}
 
-    for check in checks:
-        f.addStep(LitTestCommand(name=trunc50("test-%s-%s" % (step_name, check)),
-                                 command=['ninja', check],
-                                 description=[
-                                   "Test", "just", "built", "components", "for",
-                                   check,
-                                 ],
-                                 env=check_env,
-                                 workdir=obj_dir,
-                                 **kwargs # Pass through all the extra arguments.
-                                 ))
+    if checks:
+        for check in checks:
+            f.addStep(LitTestCommand(name=trunc50("test-%s-%s" % (step_name, check)),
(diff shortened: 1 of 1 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -2990,7 +2990,6 @@
                     bolttests=False,
                     depends_on_projects=['bolt', 'lld', 'llvm'],
                     extra_configure_args=[
-                        "-DLLVM_APPEND_VC_REV=OFF",
```
aaupov: This option is aligned with nfc builder to improve ccache hit rate. Unless there's a reason to drop this, I'd rather keep it.

buildbot/osuosl/master/config/builders.py
```diff
@@ -3018,53 +3016,6 @@
                         ],
                     )},
 
-    {'name': "bolt-aarch64-ubuntu-clang",
```
aaupov, line 3021: We've decided not to drop these builders, right?

paschalis-mpeis: Correct. That is a Stacked PR, using [user-branches](https://llvm.org/docs/GitHub.html#stacked-pull-requests), and I have to rebase it to pick up the latest changes. Once I do that would be gone.

buildbot/osuosl/master/config/builders.py
paschalis-mpeis: Agreed, reinstated.

paschalis-mpeis:
Forced push to rebase on the PR stack. Also addressed comments.

aaupov (approved):
Thanks for the cleanup.

PR #470, opened 2025-06-09
Add arm qemu libc buildbot config

amykhuang:
We will be replacing the physical arm32 libc buildbot with an x86_64 buildbot using qemu.

### DIFF
--- a/zorg/buildbot/builders/annotated/libc-linux.py
+++ b/zorg/buildbot/builders/annotated/libc-linux.py
@@ -30,6 +30,12 @@ def is_x86_64_builder(builder_name):
 def is_riscv32_builder(builder_name):
     return 'riscv32' in builder_name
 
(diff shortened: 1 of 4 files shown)


zorg/buildbot/builders/annotated/libc-linux.py
```diff
@@ -30,6 +30,9 @@ def is_x86_64_builder(builder_name):
 def is_riscv32_builder(builder_name):
     return 'riscv32' in builder_name
 
+def is_arm32_qemu_builder(builder_name):
+    return 'arm32' in builder_name and 'qemu' in builder_name
```
michaelrj-google: if you're planning to make more than just the `arm32` builder using `qemu` it might be better to have separate `qemu` and `arm32` flags.

PR #474, opened 2025-06-17
[llvm-zorg] Set up libc++ premerge testing runner sets.

cmtice:
Create new runner sets for  libc++ premerge testing in the LLVM Github premerge testing infrastructure.


cmtice:
There's probably a lot I've done wrong -- I'm considering this a learning exercise...

premerge/gke_cluster/main.tf
```diff
@@ -65,6 +65,38 @@ resource "google_container_node_pool" "llvm_premerge_linux" {
...
+  initial_node_count = 0
+
+  autoscaling {
+    total_min_node_count = 0
+    total_max_node_count = 8
+  }
+
+  node_config {
+    machine_type = var.linux_machine_type
+    taint {
```
boomanaiden154, line 81: We probably want an additional taint here specifying that this is specifically the libcxx node pool so that other jobs don't schedule here.

We could also rename the taints for the existing node pools to be specific to the main premerge system and then make this one specific to libcxx.

premerge/gke_cluster/main.tf
```diff
@@ -65,6 +65,38 @@ resource "google_container_node_pool" "llvm_premerge_linux" {
...
+  cluster            = google_container_cluster.llvm_premerge.name
+  initial_node_count = 0
+
+  autoscaling {
+    total_min_node_count = 0
+    total_max_node_count = 8
+  }
+
+  node_config {
+    machine_type = var.linux_machine_type
```
boomanaiden154: We probably want to change this to a different parameter specific to the libcxx jobs given we want to use a smaller machine shape.

premerge/premerge_resources/main.tf
```diff
@@ -111,6 +180,60 @@ resource "helm_release" "github_actions_runner_set_windows" {
...
 
+resource "helm_release" "github_actions_runner_set_libcxx" {
+  name       = "llvm-premerge-libcxx-runners"
+  namespace  = "llvm-premerge-libcxx-runners"
+  repository = "oci://ghcr.io/actions/actions-runner-controller-charts"
+  version    = "0.11.0"
+  chart      = "gha-runner-scale-set"
+
+  values = [
+    "${templatefile("libcxx_runners_values.yaml", { runner_group_name : var.runner_group_name })}"
```
boomanaiden154: You should be able to pull the container from `libcxx_image_values.yaml` here too and templatize it. That would also mean we only need one runner_values.yaml for all three configurations.

premerge/premerge_resources/main.tf
```diff
@@ -40,10 +64,55 @@ resource "kubernetes_secret" "linux_github_pat" {
   depends_on = [kubernetes_namespace.llvm_premerge_linux_runners]
 }
 
-resource "kubernetes_namespace" "llvm_premerge_windows_runners" {
+resource "kubernetes_secret" "libcxx_github_pat" {
```
boomanaiden154, line 67: We shouldn't need additional tokens. We can reuse the ones we already have.

premerge/libcxx_runners_values.yaml
```diff
@@ -0,0 +1,54 @@
...
+        #    A value lower than that could allow the scheduler to put 2
+        #    runners on the same node. Meaning 2 jobs sharing the resources of
+        #    a single node.
+        #  - lower than number_of_cores:
+        #    Each pod has some basic services running (metrics for ex). Those
+        #    already require some amount of CPU (~0.5). This means we don't
+        #    exactly have N cores to allocate, but N - epsilon.
+        #
+        # We also need to request sufficient memory to not get OOM killed.
+        requests:
```
boomanaiden154, line 41: This need to be updated based on the new machine shape.

boomanaiden154 (commented):
Some initial comments.

premerge/gke_cluster/main.tf
cmtice: Do you mean I should keep this taint and add a second one? or I should only have the one taint here, but update it? 

re renaming taints:  I assume that means I would change the value for the "key"?  The current value is 'premerge-platform', which is then also used as a key for the labels field below. If I change the taint to something like ' key = "foo"' (i'd choose a better name), then I would need to change labels to something like 'labels = { "foo" : "linux" }'? And then I would also need to change  the tolerations key and nodeSelector in the appropriate *_runners_values.yaml file(s), changing 'premerge-platform' to 'foo'?

premerge/gke_cluster/main.tf
cmtice: I don't understand how we can use  "a smaller machine shape"?  We've already requested and received the capacity increase:1024 n2d cores in us-central-1a and 1024 n2d cores in us-west1.  What "machine shape" could/would we specify?

premerge/premerge_resources/main.tf
cmtice: Could you elaborate on what you're suggesting here? I'm not sure I understand. (BTW, I have not found ANY good documentation that explains either the terraform or the yaml commands and configurations, so I've been staring at your code and guessing a lot...do you have links to any useful references?)

premerge/libcxx_runners_values.yaml
cmtice: Again, what is the new machine shape? Where/how do we define it? How do we figure out what it should be?

premerge/gke_cluster/main.tf
boomanaiden154: Machine shape as in the number of cores/amount of RAM in an individual instance. Instead of setting `libcxx_machine_type` to something like `n2d-standard-64`, we should probably be using something smaller. This gives us more instances with less cores which is better for thorughput.

Given the existing instances are 32 cores, starting with `n2d-standard-32` and `n2-standard-32` (depending upon what quota is available) probably makes the most sense.

premerge/premerge_resources/main.tf
boomanaiden154: https://developer.hashicorp.com/terraform/language/functions/templatefile should explain how the template files work.

It's probably easiest if we just take the contents of `libcxx_image_values` and put them in the terraform file. The you can combine the  `libcxx_*_runnvers_values.yaml` into a single yaml file that is templatized using the syntax above. We then just pass the images, runner set names and whatever else is different between them as template parameters.

premerge/gke_cluster/main.tf
boomanaiden154: You should be able to do either.

I would suggest just changing the `premerge-platform` label value from something like `linux` to `linux-libcxx` along with the relevant updates in the runner values YAML.

cmtice:
I have tried to update this as requested; please take another look.

premerge/libcxx_runners_values.yaml
boomanaiden154: These still need to be updated. I would set the requests to 28 on the CPU side and 100Gi on the memory side. For the limits I would do 32 and 128Gi to match the machine shape (`n2-standard-32`/`n2d-standard-32`).

premerge/premerge_resources/main.tf
```diff
@@ -111,6 +129,60 @@ resource "helm_release" "github_actions_runner_set_windows" {
...
+  chart      = "gha-runner-scale-set"
+
+  values = [
+    "${templatefile("libcxx_runners_values.yaml", { runner_group_name : var.runner_group_name, runner_image : var.libcxx_runner_image })}"
+  ]
+
+  depends_on = [
+    kubernetes_namespace.llvm_premerge_libcxx_runners,
+    helm_release.github_actions_runner_controller,
+    kubernetes_secret.linux_github_pat,
```
boomanaiden154: My earlier comment probably was pretty vague about the secrets (or I was misreading things), but we do need kubernetes secrets for each of the new namespaces as secrets in k8s are namespaced. You can essentially just copy and paste the definition of `kubernetes_secret.linux_github_pat`, adjusting the namespaces as necessary.

premerge/gke_cluster/main.tf
```diff
@@ -65,6 +65,38 @@ resource "google_container_node_pool" "llvm_premerge_linux" {
...
 
+resource "google_container_node_pool" "llvm_premerge_libcxx" {
+  name               = "llvm-premerge-libcxx"
+  location           = var.region
+  cluster            = google_container_cluster.llvm_premerge.name
+  initial_node_count = 0
+
+  autoscaling {
+    total_min_node_count = 0
+    total_max_node_count = 8
```
boomanaiden154: Now that we have 1024 cores and 32 core instances, we can bump this up to 32.

premerge/libcxx_runners_values.yaml
```diff
@@ -0,0 +1,47 @@
+githubConfigUrl: "https://github.com/llvm"
+githubConfigSecret: "github-token"
+
+minRunners: 0
+maxRunners: 8
```
boomanaiden154: This can get bumped up to 32 as well.

Here there is a possibility that we run into some issues if both the new runner set and the older runner set try to spawn a total of more than 16 jobs at the same time (pods would queue), but theoretically things should just queue. In practice we might run into some minor issues based on what I've observed, but nothing major, and the switchovers occur pretty rarely anyways.

boomanaiden154 (commented):
Almost looks good.

premerge/premerge_resources/main.tf
cmtice: Done.

premerge/libcxx_runners_values.yaml
cmtice: Done.

premerge/libcxx_runners_values.yaml
cmtice: Done.

premerge/gke_cluster/main.tf
cmtice: Done.

premerge/premerge_resources/variables.tf
```diff
@@ -54,3 +54,18 @@ variable "externalservices_tempo_basicauth_username" {
...
+}
+
+variable "libcxx_release_runner_image" {
+  type = string
+  default = "ghcr.io/llvm/libcxx-linux-builder:d8a0709b1090350a7fe3604d8ab78c7d62f10698"
+}
+
+variable "libcxx_next_runner_image" {
+  type = string
+  default = ""
```
boomanaiden154: This should probably be filled out with something, even if it's just copied from `libcxx_runner_image` for now.

cmtice: Done

cmtice:
> Could you put the output of running `terraform plan` into a Github Gist/paste somewhere so we can see that there's nothing weird going on before you run `terraform apply`?

I have put the output from 'terraform plan' in https://gist.github.com/cmtice/ef54aa9100e4fa144379fb4c137c26a5

Let me know if you can't see it.

PR #476, opened 2025-06-18
Fix test/jenkins/test_build.py

cachemeifyoucan:
Update the test after updating the build script configuration


cachemeifyoucan:
I see. The lit test is XFAILed in automation test so it never caught the failure. It passed locally for me. Let me merge this to make sure it passes in automation.

PR #479, opened 2025-06-19
[RISCV] Add additional RISC-V 2-stage cross-compile and test under qemu-system configs

asb:
This adds:
* rva23 with zvl512b
* rva23 with zvl1024b
* -mcpu=spacemit-x60 with -mrvv-vector-bits=zvl (this config also gives zvl256b coverage)


lukel97 (approved):
LGTM

PR #480, opened 2025-06-20
[BOLT] Improve handling in BOLTBuilder and cleanup

paschalis-mpeis:
- Add a clean-nfc-check step to remove auxiliary files before nfc-mode-setup.
- Escalate errors in nfc-check-setup.py as warnings. Checks continue as expected.
- Allow tests to proceed if the old/new binaries are not generated. (rare case)
- Provide clearer message when GNU build IDs are embedded in the binary but they do match. Tests are skip accordingly.

### DIFF
--- a/zorg/buildbot/builders/BOLTBuilder.py
+++ b/zorg/buildbot/builders/BOLTBuilder.py
@@ -99,32 +99,49 @@ def getBOLTCmakeBuildFactory(
         # Individual markers to skip either in-tree or out-of-tree tests.
         skipInTree = ".llvm-bolt.skip.in-tree"
         skipOutOfTree = ".llvm-bolt.skip.out-of-tree"
+        boltNew = "bin/llvm-bolt.new"
+        boltOld = "bin/llvm-bolt.old"
 
         f.addSteps([
             ShellCommand(
+                name='clean-nfc-check',
+                command=(
+                        f"rm -f {boltNew} {boltOld} {hasSrcChanges} {skipInTree} {skipOutOfTree}; "
+                        ),
+                description=('Cleanup for NFC-Mode testing'),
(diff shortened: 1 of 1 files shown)


zorg/buildbot/builders/BOLTBuilder.py
```diff
@@ -137,14 +154,14 @@ def getBOLTCmakeBuildFactory(
...
                 name='nfc-check-bolt-different',
                 command=(
-                          f'rm -f {skipInTree} {skipOutOfTree}; '
-                          f'cmp -s bin/llvm-bolt.old bin/llvm-bolt.new && ('
+                          f'cmp -s {boltNew} {boltOld} && ('
                           f'touch {skipInTree}; touch {skipOutOfTree}); '
-                          f'[ -f {hasSrcChanges} ] && rm -f {skipInTree};'
-                          f'rm -f {hasSrcChanges}; '
+                          f'[ -f {hasSrcChanges} ] && rm -f {skipInTree}; '
+                          f'return 0'
```
aaupov, line 169: Do you need to return 0 explicitly?

paschalis-mpeis: Yes, but it's subtle. When the condition is:
1. **true:** (ie, `hasSrcChanges` marker is there), we proceed with in-tree tests. And `rm -f` returns `0` ✅ 
2. **false:** the exit code is `1`, which is escalated as a warning ❌ 



---
So this is needed for case 2 to avoid:

Screenshot 2025-06-23 at 13 42 55



> program finished with **exit code 1**

aaupov: Thank you for explaining.

PR #482, opened 2025-06-24
[flang] Add dependency for out-of-tree flang on compiler-rt

DavidTruby:
flang on AArch64 now always depends on compiler-rt so it needs to be
built in the main LLVM build for the out-of-tree flang bots.

### DIFF
--- a/zorg/buildbot/builders/FlangBuilder.py
+++ b/zorg/buildbot/builders/FlangBuilder.py
@@ -19,7 +19,7 @@ def getFlangOutOfTreeBuildFactory(
     f = getCmakeWithNinjaBuildFactory(
             depends_on_projects=['llvm','clang','mlir','openmp','flang','flang-rt'],
             enable_projects=['llvm','clang','mlir'],
-            enable_runtimes=['openmp'],
+            enable_runtimes=['openmp', 'compiler-rt'],
             obj_dir="build_llvm",
             checks=[],
             clean=clean,


zorg/buildbot/builders/FlangBuilder.py
```diff
@@ -19,7 +19,7 @@ def getFlangOutOfTreeBuildFactory(
     f = getCmakeWithNinjaBuildFactory(
             depends_on_projects=['llvm','clang','mlir','openmp','flang','flang-rt'],
             enable_projects=['llvm','clang','mlir'],
-            enable_runtimes=['openmp'],
+            enable_runtimes=['openmp', 'compiler-rt'],
```
luporl: I'm not sure, but I think that 'compiler-rt' should also be added to `depends_on_projects`.
My understanding is that this way builds are also triggered by changes in compiler-rt.

DavidSpickett:
https://lab.llvm.org/staging/#/builders/46/builds/4084 uses the out of tree config now too, so this will fix that builder as well.

I don't see any more of our Flang builders with this issue, so we can assume they don't need changing.

DavidSpickett (approved):
LGTM

DavidTruby:
In-tree flang has the dependency listed in the cmake so compiler-rt will get auto added, in the same way flang-rt does. There's just no real way to do this for out of tree flang unfortunately.

DavidSpickett:
This is now present on staging, but the flang-libcxx builder is still failing the test.

The initial llvm build has compiler-rt now:
```
cmake -DLLVM_TARGETS_TO_BUILD=AArch64 -DCMAKE_CXX_STANDARD=17 -DLLVM_ENABLE_WERROR=OFF -DLLVM_ENABLE_ASSERTIONS=ON -DLLVM_ENABLE_LIBCXX=On -DCMAKE_BUILD_TYPE=Release '-DLLVM_ENABLE_PROJECTS=llvm;clang;mlir' '-DLLVM_ENABLE_RUNTIMES=openmp;compiler-rt' '-DLLVM_LIT_ARGS=-v -vv' -GNinja ../llvm-project/llvm
```
Test fails:
```
/usr/bin/ld: cannot find /home/tcwg-buildbot/worker/flang-aarch64-libcxx/build_flang/lib/clang/21/lib/aarch64-unknown-linux-gnu/libclang_rt.builtins.a: No such file or directory
flang-21: error: linker command failed with exit code 1 (use -v to see invocation)
```
I see two of those files, but not in that location:
```
/worker/flang-aarch64-libcxx$ find -name "*.a" | grep aarch64 | grep builtins | grep -v llvm-project
./build/lib/clang/21/lib/aarch64-unknown-linux-gnu/libclang_rt.builtins.a
./build_llvm/lib/clang/21/lib/aarch64-unknown-linux-gnu/libclang_rt.builtins.a
```
I think `./build` is just the old build directory from before the configuration change. `./build_llvm` is the current one.

@luporl can you take a look at this?

DavidTruby:
Sorry, I just got around to looking at this again; clang has the same issue as flang here if built out of tree with -DCLANG_DEFAULT_RTLIB=compiler-rt. The driver code simply assumes that the path to compiler-rt is relative to the driver itself. I think changing this might be non-trivial.
Could we possibly fix this with CMAKE_INSTALL_PREFIX being the same for both the LLVM build and the flang build, and installing both there? I guess you can't run the tests that way...

DavidTruby:
I think @MaskRay might know better what the logic is for how clang looks for compiler-rt; maybe they have an idea what we could do?

DavidTruby:
I have tried it without zorg and it seems to work. Fingers crossed!
That would at least get these bots passing again; I think the underlying issue still needs resolving though so I've opened an llvm-project issue for it.

PR #485, opened 2025-06-26
Remove PSTL from the monolithic builders

AaronBallman:
This directory was removed in
https://github.com/llvm/llvm-project/commit/f8ed45611b75ecac1d459aa6ae3b7051d435abe6


gkistanova:
No restart is needed in this case. You should be good with this change.
Anyway, the restart is coming tonight.

AaronBallman:
> No restart is needed in this case. You should be good with this change. Anyway, the restart is coming tonight.

Thanks! For my own education, then why did landing the changes to remove PSTL from the CMake not actually address the issue?

I landed my changes long before @rnk landed his workaround in: https://lab.llvm.org/buildbot/#/builders/153/builds/36237 which is when the bot went back to green.

PR #487, opened 2025-06-28
[BOLT] Add sanity check for frozen llvm-bolt

paschalis-mpeis:
Some patches can cause the llvm-bolt binary to hang, which stalls or fails the test pipeline.

Add a simple sanity check that runs:
```
llvm-bolt --version
```

with a 30-second timeout. If the command does not complete in time, flunk the build.

Also, sets maxTime for nfc-check-validation and reduce the number of lit workers for in-tree tests.

### DIFF
--- a/zorg/buildbot/builders/BOLTBuilder.py
+++ b/zorg/buildbot/builders/BOLTBuilder.py
@@ -206,7 +206,7 @@ def getBOLTCmakeBuildFactory(
             # Lower scheduling priority, as above.
             LitTestCommand(
                 name='nfc-check-large-bolt',
-                command=('nice -n 5 bin/llvm-lit -sv -j2 tools/bolttests'),
+                command=('nice -n 5 bin/llvm-lit -v -j2 tools/bolttests'),
                 description=["running", "NFC", "check-large-bolt"],
                 descriptionDone=["NFC", "check-large-bolt", "completed"],
                 warnOnFailure=True,
--- a/zorg/buildbot/builders/BOLTBuilder.py
+++ b/zorg/buildbot/builders/BOLTBuilder.py
@@ -130,6 +130,8 @@ def getBOLTCmakeBuildFactory(
                 haltOnFailure=False,
                 flunkOnFailure=False,
                 env=env),
+            # Verify that the llvm-bolt binary can report its version within a
+            # reasonable amount of time.
             ShellCommand(
                 name='llvm-bolt-version-check',
                 command=(f"{boltNew} --version"),
@@ -164,7 +166,7 @@ def getBOLTCmakeBuildFactory(
                 haltOnFailure=False,
                 warnOnFailure=True,
                 warnOnWarnings=True,
-                maxTime=20,
+                maxTime=30,
                 decodeRC={0: SUCCESS, 1: FAILURE, 2: WARNINGS},
                 descriptionDone=["NFC-Mode Validation"],
                 env=env),
@@ -185,10 +187,12 @@ def getBOLTCmakeBuildFactory(
                 haltOnFailure=False,
                 env=env),
             # Run in-tree tests if the llvm-bolt binary has changed, or if
-            # relevant source code changes are detected.
+            # relevant source code changes are detected. Lower scheduling
+            # priority with nice to reduce CPU contention in virtualized
+            # environments.
             LitTestCommand(
                 name='nfc-check-bolt',
-                command=("LIT_OPTS='-j2' nice -n 5 ninja check-bolt"),
+                command=("nice -n 5 ninja check-bolt"),
                 description=["running", "NFC", "check-bolt"],
                 descriptionDone=["NFC", "check-bolt", "completed"],
                 warnOnFailure=True,
@@ -197,6 +201,7 @@ def getBOLTCmakeBuildFactory(
                 doStepIf=FileDoesNotExist(f"build/{skipInTree}"),
                 env=env),
             # Run out-of-tree large tests if the llvm-bolt binary has changed.
+            # Lower scheduling priority, as above.
             LitTestCommand(
                 name='nfc-check-large-bolt',
                 command=('nice -n 5 bin/llvm-lit -sv -j2 tools/bolttests'),
--- a/zorg/buildbot/builders/BOLTBuilder.py
+++ b/zorg/buildbot/builders/BOLTBuilder.py
@@ -188,7 +188,7 @@ def getBOLTCmakeBuildFactory(
             # relevant source code changes are detected.
             LitTestCommand(
                 name='nfc-check-bolt',
-                command=("LIT_OPTS='-j2' ninja check-bolt"),
+                command=("LIT_OPTS='-j2' nice -n 5 ninja check-bolt"),
                 description=["running", "NFC", "check-bolt"],
                 descriptionDone=["NFC", "check-bolt", "completed"],
                 warnOnFailure=True,
@@ -199,8 +199,7 @@ def getBOLTCmakeBuildFactory(
             # Run out-of-tree large tests if the llvm-bolt binary has changed.
             LitTestCommand(
                 name='nfc-check-large-bolt',
-                command=['bin/llvm-lit', '-sv', '-j2',
-                         'tools/bolttests'],
+                command=('nice -n 5 bin/llvm-lit -sv -j2 tools/bolttests'),
                 description=["running", "NFC", "check-large-bolt"],
                 descriptionDone=["NFC", "check-large-bolt", "completed"],
                 warnOnFailure=True,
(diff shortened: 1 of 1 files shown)


paschalis-mpeis:
It looks like Buildbot stalls on a [particular patch](https://lab.llvm.org/staging/#/builders/219/builds/121) that freezes `llvm-bolt`, causing the worker process to panic.

I limited the lit workers for `nfc-check-bolt` to two, but both this step, as well the original `nfc-check-large-bolt`, still struggle. At best, they take TIMEOUT/WORKERS to complete.

With this patch, I detect hangs early and bail out cleanly.
Any further advice would be appreciated.

paschalis-mpeis:
TL;DR: Not an incremental build failure, but leftover llvm-project wrapper scripts from the previous NFC-Mode logic that I should have communicated and cleaned up.


## A) Why this happens:
NFC-Mode runs tests only when the `llvm-bolt` binary changes between revisions, or when relevant files (under `bolt/`) are modified.

### How NFC-Mode is set up:

1. **build-bolt** step: compiles the `llvm-bolt` binary.

2. **nfc-check-setup** step:
    - renames the current `llvm-bolt` binary to `llvm-bolt.new`. 
    - Builds the previous SHA into `llvm-bolt.old`.
    - Creates a helper python wrapper `llvm-bolt` that redirects to `llvm-bolt.new`

On disk:
- `llvm-bolt` → `llvm-bolt.new` (python wrapper)
- `llvm-bolt.new` (current SHA binary)
- `llvm-bolt.old` (previous SHA binary)

3. **nfc-check-bolt** step: builds test utilities and runs tests, with `check-bolt`, which relinks `llvm-bolt` and overwrites the wrapper.

### Reproducing the problem
(1) Buildbot runs https://github.com/llvm/llvm-project/pull/146148 which modifies the python wrapper script

(2) After PR (1), buildbot runs any patch that does not modify llvm-bolt, like: https://github.com/llvm/llvm-project/pull/145812

Under the updated [ninja](https://github.com/llvm/llvm-zorg/pull/466) flow, it results on disk:
- `llvm-bolt` → `llvm-bolt.new` (python wrapper)
- `llvm-bolt.new` → itself (python wrapper **recursion**) 🔁 
- `llvm-bolt.old` (previous SHA binary)

This creates an infinite recursion.

## B) Proposed solution:

Since BOLTBuilder no longer requires a wrapper for `llvm-bolt`, we can retire that logic with [#146209](https://github.com/llvm/llvm-project/pull/146209). @aaupov, do you agree?

This change would consistently avoid the above scenario. I tested it on an internal buildbot with a manual local edit of [#146209](https://github.com/llvm/llvm-project/pull/146209) and running `git update-index --assume-unchanged bolt/utils/nfc-check-setup.py`. Now repeating on the staging buildbot; so far works as expected.


## C) Is this PR needed?
If we remove the wrapper, this patch becomes optional. It can serve as a fail-safe if `llvm-bolt` loops (for some reason). Tuning niceness for vCPU is not a bad idea, and the extra step won't hurt. I can still increase the lit workers though.

DavidSpickett:
I have no knowledge of the previous setup, but I can say that if someone committed a patch that made any use of clang loop forever, every existing bot would timeout the thousands of clang tests. Might hit the overall test step timeout, but might not if it printed something once in a while.

My point with that is - this NFC idea sounds nice but it's adding a lot of complexity. Even if it's not as complex as I think, the fact that I think it is, is at least cognitive complexity.

(but I have literally just read this PR and your explanation, so definitely could be a skill issue on my part)

So not having it would be no worse than the other bots. You can decide if that's a good or a bad thing.

Is Bolt particularly prone to these timeouts? It's always theoretically possible with clang but never happened enough to try to prevent it.

If bolt experts are cool with this or the other patch, great. If undecided maybe you can explain more what you're doing and I'll try to give an outsider's perspective on the implementation.

DavidSpickett:
Myself and @paschalis-mpeis discussed this and I understand better the sequence of events here.

* Bolt has an NFC mode that allows the builder to skip running expensive external tests.
  * This is very cool and worth keeping IMO.
* Even if the bolt binary did not change, if the change set has changes to Bolt tests, they should be re-run at least.
* `ninja <the target for doing that>` may try to re-link or rebuild the bolt binary, except that it's not a binary, it's a script to implement this NFC mode, so that script would get replaced.
* This situation occurred recently and motivated this PR to mitigate the impact when that does happen.
* In future, @paschalis-mpeis wants to improve the NFC mode to avoid this situation completely.
* If the NFC mode no longer has this flaw, this PR is not needed to prevent this specific situation.
* However it would mitigate someone causing bolt to stall by actually changing bolt.

This is not a check we do on the other bots for clang etc., but that's just because we've never had anyone (intentionally or otherwise) cause a widespread problem. If they do in future, maybe we will go down this road too.

And if no one ever makes bolt stall, then great! We just spend a few cycles in case.

zorg/buildbot/builders/BOLTBuilder.py
```diff
@@ -130,6 +130,16 @@ def getBOLTCmakeBuildFactory(
                 haltOnFailure=False,
                 flunkOnFailure=False,
                 env=env),
+            ShellCommand(
+                name='llvm-bolt-version-check',
+                command=(f"{boltNew} --version"),
+                description=('Check that llvm-bolt binary passes a simple test'
```
DavidSpickett, line 140: The alternative would be some sort of dependency system in the test suite itself where you could say "must be able to print version" comes before "must be able to print help" comes before....and so on. This is not available in lit today, and is basically a build system unto itself. Fun idea, bad ROI.

So a smoke test like this is fine.

zorg/buildbot/builders/BOLTBuilder.py
```diff
@@ -130,6 +130,16 @@ def getBOLTCmakeBuildFactory(
...
+                             'before proceeding with testing.'),
+                descriptionDone=["llvm-bolt --version"],
+                haltOnFailure=True,
+                flunkOnFailure=True,
+                maxTime=30,
```
DavidSpickett, line 145: This part has the whole worker to itself doesn't it?

(just confirming because time limits are ineffective when you have lots of processes fighting for resources)

zorg/buildbot/builders/BOLTBuilder.py
```diff
@@ -154,6 +164,7 @@ def getBOLTCmakeBuildFactory(
                 haltOnFailure=False,
                 warnOnFailure=True,
                 warnOnWarnings=True,
+                maxTime=20,
```
DavidSpickett: This step may invoke the wrapper script, that's why you need a time limit here too, right?

zorg/buildbot/builders/BOLTBuilder.py
```diff
@@ -177,7 +188,7 @@ def getBOLTCmakeBuildFactory(
             # relevant source code changes are detected.
             LitTestCommand(
                 name='nfc-check-bolt',
-                command=["ninja", "check-bolt"],
+                command=("LIT_OPTS='-j2' nice -n 5 ninja check-bolt"),
```
DavidSpickett: You said you might remove the -j and I agree with doing that, but if you didn't want to, there is a cmake option `LLVM_LIT_ARGS` you can set that will apply to everything (can be overridden by env vars and local config)

DavidSpickett: And a comment here to explain the use of nice would be......nice. :drum:

zorg/buildbot/builders/BOLTBuilder.py
paschalis-mpeis: Correct, the worker processes one job at a time.

zorg/buildbot/builders/BOLTBuilder.py
paschalis-mpeis: Correct. Probably I should also use 30 seconds here for consistency.

The previous step shouldn't let us reach this point. If it's agreed and the below PR goes ahead, it won't be needed unless something catastrophic makes it into production code:
- https://github.com/llvm/llvm-project/pull/146209

zorg/buildbot/builders/BOLTBuilder.py
paschalis-mpeis: Thanks for the tips, David.

Explaining the use of nice would be nice indeed. 🙂 
I will send a follow-up patch, adding some comments to the code.

paschalis-mpeis: Added some comments.
Under heavy resource contention, `nice` alone may not be enough. Still better than not using it.

paschalis-mpeis:
Hey David,

Thanks for your review. Please let me know of any further comments

aaupov (approved):

PR #488, opened 2025-06-30
Enable OpenMP and Add new WoA 2stage worker

omjavaid:
This patch introduces the following changes:

1. A new AArch64 Windows worker linaro-armv8-windows-msvc-03 for clang-arm64-windows-msvc-2stage builder.

3. The OpenMP is enabled for the clang-arm64-windows-msvc-2stage builder via -DLLVM_ENABLE_RUNTIMES=openmp CMake flag.


DavidSpickett (commented):
Can you state whether you intend this to run tests or just enable it so flang can use it?

If it's the former, I assume the tests have been stable? OpenMP likes parallelism so it's a risk.

PR #491, opened 2025-07-09
[Buildbot][AMD] Make builder script to take argument

Kewen12:
Updated the annotated builder so that it takes cmake filename as the argument. With the default value, this change should not disrupt current runs. We will make incremental changes in the next few PRs.


Kewen12:
Please help review this change @jhuber6 @jplehr  Thanks!

PR #494, opened 2025-07-10
[CI] Move ops container under premerge

boomanaiden154:
Top level directories should probably be reserved for majorish efforts. I think the ops-container definitions make a lot more sense to have under the premerge folder even if they are not perfectly aligned.


boomanaiden154:
@jriv01

cmtice:
I just noticed that this PR is still unreviewed.  Is this work still something you want to commit? Should I review it or do you just want to close it?

boomanaiden154:
> I just noticed that this PR is still unreviewed. Is this work still something you want to commit? Should I review it or do you just want to close it?

I think this is still something we want to do.

jriv01 (approved):

PR #500, opened 2025-07-15
Fix out-of-tree Flang builders

luporl:
Currently, when Flang is built out-of-tree, it fails to find
compiler-rt when linking programs. This issue was reported in
llvm/llvm-project#147738 and a fix for it is being discussed.

Meanwhile, there is a workaround available, discussed in #482, that
consists in using an installed flang instead of the one in the
build directory. This PR changes Flang builder to use this
workaround.

### DIFF
--- a/zorg/buildbot/builders/FlangBuilder.py
+++ b/zorg/buildbot/builders/FlangBuilder.py
@@ -16,6 +16,11 @@ def getFlangOutOfTreeBuildFactory(
     if env is None:
         env = dict()
 
+    # Currently, when Flang is built out-of-tree, it fails to find compiler-rt
+    # when linking programs.
+    # See https://github.com/llvm/llvm-project/issues/147738.
+    # As a workaround, install LLVM, clang, compiler-rt and flang into the same
+    # directory and use it to build and test flang-rt.
     install_dir = "install"
 
     f = getCmakeWithNinjaBuildFactory(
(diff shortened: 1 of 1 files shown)


zorg/buildbot/builders/FlangBuilder.py
```diff
@@ -16,11 +16,14 @@ def getFlangOutOfTreeBuildFactory(
     if env is None:
         env = dict()
 
+    install_dir = "install"
```
DavidSpickett, line 24: I assume that by passing install_dir to the factory function below, that generates an install step. If you leave it out, it does not.

Can you put a comment here that links to the issue we're working around?

Meinersbur: I'd also add the info that the `getCmakeWithNinjaBuildFactory` and `addNinjaSteps` later are installing the files into a common install_dir/prefix combining clang's/flang's resource directory

luporl: Yes, this assumption is correct, by passing install_dir `getCmakeWithNinjaBuildFactory` and `addNinjaSteps`  generate an install step. The factory function also takes care of cleaning it.

I have linked the issue and documented the workaround. Please let me know if it's still not clear.

DavidTruby:
These still seem to be failing on the main buildbot instance; is that still pending a restart or something? Is there any way we can trigger that or does it just happen periodically?

PR #503, opened 2025-07-17
[Github] Add Terraform Formatting Check

boomanaiden154:
This patch adds a terraform formatting check to the llvm zorg CI. This just
makes it easier to verify that patches have been formatted before landing which
prevents the need for follow up commits to reformat things later.


.github/workflows/tf-formatting.yml
```diff
@@ -0,0 +1,29 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y curl
+        curl https://releases.hashicorp.com/terraform/1.12.2/terraform_1.12.2_linux_amd64.zip > /tmp/tf.zip
+        cd /tmp
+        unzip /tmp/tf.zip
+        cp terraform /usr/local/bin/terraform
+    - name: Check Formatting
+      run: |
+        terraform fmt -recursive .
```
cmtice: Shouldn't you add the "--check" flag? I think, as written, this will actually do the reformatting, rather than just checking to see if the format is valid.

boomanaiden154: Ah, yep. I was experimenting with that in my terminal and then apparently forgot to actually update the workflow file.

Thank you for pointing this out.

.github/workflows/tf-formatting.yml
```diff
@@ -0,0 +1,29 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y curl
+        curl https://releases.hashicorp.com/terraform/1.12.2/terraform_1.12.2_linux_amd64.zip > /tmp/tf.zip
+        cd /tmp
+        unzip /tmp/tf.zip
+        cp terraform /usr/local/bin/terraform
+    - name: Check Formatting
+      run: |
+        terraform fmt -recursive --check .
```
cmtice: According to terraform fmt --help, this option should only have one hyphen,.

boomanaiden154: It appears to work either way. Updated to one hyphen for consistency with the `-recursive` flag.

PR #506, opened 2025-07-18
[CI] Enable Workload Identity Federation on Linux/Windows Nodepools

boomanaiden154:
This patch enables workload identity federation on the Linux and Windows node
pools as described in
https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity. This
is necessary for authenticating with the GCS APIs so that we can use sccache
with a GCS bucket for caching.


premerge/gke_cluster/main.tf
```diff
@@ -62,6 +69,12 @@ resource "google_container_node_pool" "llvm_premerge_linux" {
...
       "goog-gke-node-pool-provisioning-model" = "on-demand"
     }
+
+    # Enable workload identity federation for this pool so that we can access
+    # GCS buckets.
+    workload_metadata_config {
+      mode = "GKE_METADATA"
+    }
   }
 }
```
cmtice, line 80: Is there any particular reason why we're not adding this to the llvm_premerge_libcxx node pool?

boomanaiden154: We're not planning on using `sccache` for them given the libcxx build is really quick, so we don't need to pass any authentication tokens around. The expensive part for the libcxx CI is the test suite which we can't cache.

cmtice (approved):

PR #507, opened 2025-07-18
[CI] Add namespace names as parameters to premerge_resources

boomanaiden154:
This enables a bit of simplification as we can remove some strings where they
really should be variables. This will also help in a future patch where we also
need to access the namespace name inside the gke_cluster module as it will
enable having a common definition to ensure that they do not differ.


premerge/premerge_resources/main.tf
```diff
@@ -19,7 +19,7 @@ resource "kubernetes_namespace" "llvm_premerge_controller" {
 
 resource "kubernetes_namespace" "llvm_premerge_linux_runners" {
   metadata {
-    name = "llvm-premerge-linux-runners"
+    name = var.linux_runners_namespace_name
   }
 }
```
cmtice, line 25: Again, why are we not doing this for the libcxx runners?

boomanaiden154: The libcxx runners don't benefit really at all from caching, so they don't need any setup for this.

premerge/premerge_resources/main.tf
```diff
@@ -146,7 +146,7 @@ resource "helm_release" "github_actions_runner_controller" {
 
 resource "helm_release" "github_actions_runner_set_linux" {
   name       = "llvm-premerge-linux-runners"
```
cmtice, line 148: Don't you need to make the change here, too?

premerge/premerge_resources/main.tf
```diff
@@ -164,7 +164,7 @@ resource "helm_release" "github_actions_runner_set_linux" {
 
 resource "helm_release" "github_actions_runner_set_windows_2022" {
   name       = "llvm-premerge-windows-2022-runners"
```
cmtice, line 166: And here?

premerge/premerge_resources/main.tf
boomanaiden154: That's the name of the helm release. It happens to be the same as the namespace name, but it doesn't need to be, so it's fine to keep them separate.

premerge/premerge_resources/main.tf
boomanaiden154: Ditto.

premerge/premerge_resources/main.tf
cmtice: Got it. Thanks!

cmtice (approved):

PR #513, opened 2025-07-21
Increase timeout for clang-sparc64-linux to 1800 seconds

glaubitz:
The builds for clang-sparc64-linux currently fail with a timeout during the linking stage, so lets increase the timeout to 1800 seconds which is the timeout also used on the builder clang-solaris11-sparcv9.


glaubitz:
The SPARC T4 CPU used is rather slow for single-core performance which is why an increase makes sense.

See also the other SPARC builder for Solaris which also has 1800 seconds timeout.

gkistanova (approved):
We can increase the timeout now to move the issue out of the critical pass, and have the analyses later.

PR #515, opened 2025-07-21
Add another email to the list of notifications for my bots.

dyung:
Currently the email notifications go to my work address, but since some of the bots I manage are self-hosted, it would be useful to get notifications to a personal email so I don't have to check my work email to see them. This adds a personal email so that I can get notifications there as well of problems.


gkistanova (approved):
LGTM.

PR #521, opened 2025-07-23
Merge clang-aarch64-full-2stage with clang-aarch64-lld-2stage

DavidSpickett:
And keep the latter. Since we (Linaro) don't commit to testing ld specifically, and recommend lld anyway.

The ld bot had some extra testing that I've enabled for the lld version in this change.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -414,22 +414,34 @@ all = [
                     checkout_lld=False,
                     extra_cmake_args=["-DLLVM_TARGETS_TO_BUILD='AArch64'"])},
 
-    ## AArch64 check-all + LLD + test-suite 2-stage
+    # AArch64 2 staage build with lld, flang, compiler-rt, test-suite and SVE/SME
+    # mlir integration tests.
     {'name' : "clang-aarch64-lld-2stage",
     'tags'  : ["lld"],
     'workernames' : ["linaro-clang-aarch64-lld-2stage"],
     'builddir':"clang-aarch64-lld-2stage",
     'factory' : ClangBuilder.getClangCMakeBuildFactory(
                 clean=True,
+                checkout_flang=True,
+                checkout_lld=True,
                 useTwoStage=True,
+                testStage1=False,
                 runTestSuite=True,
+                env={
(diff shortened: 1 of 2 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -414,22 +414,34 @@
                     checkout_lld=False,
                     extra_cmake_args=["-DLLVM_TARGETS_TO_BUILD='AArch64'"])},
 
-    ## AArch64 check-all + LLD + test-suite 2-stage
+    # AArch64 2 staage build with lld, flang, compiler-rt, test-suite and SVE/SME
```
luporl:
```suggestion
    # AArch64 2 stage build with lld, flang, compiler-rt, test-suite and SVE/SME
```

buildbot/osuosl/master/config/builders.py
```diff
@@ -414,22 +414,34 @@
...
     {'name' : "clang-aarch64-lld-2stage",
     'tags'  : ["lld"],
     'workernames' : ["linaro-clang-aarch64-lld-2stage"],
     'builddir':"clang-aarch64-lld-2stage",
     'factory' : ClangBuilder.getClangCMakeBuildFactory(
                 clean=True,
+                checkout_flang=True,
+                checkout_lld=True,
                 useTwoStage=True,
+                testStage1=False,
```
luporl: Stage 1 tests are now being skipped because we have similar bots that run them, right?

buildbot/osuosl/master/config/builders.py
DavidSpickett: Fixed.

buildbot/osuosl/master/config/builders.py
DavidSpickett: No, I just copied it over thinking it was a default value put there to be more verbose.

In fact we do want to keep testing stage 1 here, so I've removed this parameter.

DavidSpickett: One could argue that aarch64 quick is the "similar bot" to aarch64 lld 2 stage, but the goal of this PR was to just port over the missing bits not make further changes.

Might come back to this later.

PR #522, opened 2025-07-24
[CI] Add maintenance window to premerge clusters

boomanaiden154:
This patch adds an explicit maintenance window to the premerge clusters. This is in response to some control plane upgrades that we recieved notice of that will prevent access to the k8s control plane for ~15 minutes which means we will not be able to start new jobs. We should have this anyways though as the current node upgrade strategy also breaks jobs.

### DIFF
--- a/premerge/gke_cluster/main.tf
+++ b/premerge/gke_cluster/main.tf
@@ -19,6 +19,16 @@ resource "google_container_cluster" "llvm_premerge" {
   workload_identity_config {
     workload_pool = "llvm-premerge-checks.svc.id.goog"
   }
+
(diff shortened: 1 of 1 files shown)


premerge/gke_cluster/main.tf
```diff
@@ -19,6 +19,16 @@ resource "google_container_cluster" "llvm_premerge" {
   workload_identity_config {
     workload_pool = "llvm-premerge-checks.svc.id.goog"
   }
+
+  # We prefer that maintenance is done on weekends betwee 02:00 and 05:00
```
cmtice: typo ("betwee")

boomanaiden154: Fixed, thanks for the catch!

PR #523, opened 2025-07-24
[CI] Attach pod disruption budgets to runner pods

boomanaiden154:
This patch adds some pod disruption budgets to runner pods that just sets the minimum number of available pods to the maximum. This ensure that the number of pods that k8s calculates can be disrupted is zero. This means that when GKE is updating the node pool, it must wait an hour before forcibly evicting the pod, giving it time to finish. Before this, when GKE wanted to upgrade a node, it would forcibly evict the pod very quickly (theoretically after the grace period which has a default of 30s) not realizing it is stateful.

### DIFF
--- a/premerge/pdb.yaml
+++ b/premerge/pdb.yaml
@@ -1,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: runner-set-pdb
+  namespace: ${ runner_set_name }
+spec:
+  minAvailable: ${ max_pod_count }
+  selector:
+    matchLabels:
+      actions.github.com/scale-set-name: ${ runner_set_name }
--- a/premerge/premerge_resources/main.tf
+++ b/premerge/premerge_resources/main.tf
@@ -258,6 +258,31 @@ resource "kubernetes_service_account" "windows_2022_object_cache_ksa" {
   depends_on = [kubernetes_namespace.llvm_premerge_windows_2022_runners]
 }
 
+resource "kubernetes_manifest" "linux_runners_pdb" {
+  manifest   = yamldecode(templatefile("pdb.yaml", { runner_set_name : "llvm-premerge-linux-runners", max_pod_count : 16 }))
+  depends_on = [kubernetes_namespace.llvm_premerge_linux_runners]
+}
+
+resource "kubernetes_manifest" "windows_2022_runners_pdb" {
+  manifest   = yamldecode(templatefile("pdb.yaml", { runner_set_name : "llvm-premerge-windows-2022-runners", max_pod_count : 16 }))
+  depends_on = [kubernetes_namespace.llvm_premerge_linux_runners]
+}
+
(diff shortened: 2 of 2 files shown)


premerge/pdb.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
```
cmtice: I would prefer a more descriptive file name than "pdb.yaml".

premerge/premerge_resources/main.tf
```diff
@@ -258,6 +258,31 @@ resource "kubernetes_service_account" "windows_2022_object_cache_ksa" {
   depends_on = [kubernetes_namespace.llvm_premerge_windows_2022_runners]
 }
 
+resource "kubernetes_manifest" "linux_runners_pdb" {
+  manifest   = yamldecode(templatefile("pdb.yaml", { runner_set_name : "llvm-premerge-linux-runners", max_pod_count : 16 }))
+  depends_on = [kubernetes_namespace.llvm_premerge_linux_runners]
+}
+
```
cmtice, line 270: If we decide to keep 'pdb' as an acronym for 'pod disruption budget' (and I can see why a short acronym is useful), maybe we could at least add a comment here stating what 'pdb' stands for? OR mentioning that the defintions below are for the Kubernetes Pod Disruption Budget (pdb). Something like that.

cmtice: I would also add a comment about how/why you chose those values, so someone reading through the file doesn't have to look for the PR to figure it out.

premerge/pdb.yaml
boomanaiden154: Updated to `pod-disruption-budget.yaml`.

premerge/premerge_resources/main.tf
boomanaiden154: I changed up the names to be `_disruption_budget` to make things more explicit. Never a very bad idea to expand out a (overloaded in the compiler world) acronym.

premerge/premerge_resources/main.tf
```diff
@@ -258,6 +258,36 @@ resource "kubernetes_service_account" "windows_2022_object_cache_ksa" {
   depends_on = [kubernetes_namespace.llvm_premerge_windows_2022_runners]
 }
 
+# We set up pod disruption budgets here. We need one per namespace and we need
+# to set the max pod count to the maximum number of runner pods that can
```
cmtice: Maybe:  change 'max pod count' to 'min pod count' in the comment above?

tatus
```diff
@@ -0,0 +1,59 @@
+[1mdiff --git a/premerge/pdb.yaml b/premerge/pdb.yaml[m
```
cmtice: Is there a typo in this file name ("tatus")?

boomanaiden154: The file shouldn't exist. Looks like I was typing in the wrong spot when editing the commit message in `vim` and some version of the commit ended up in a file...

cmtice (approved):

PR #526, opened 2025-07-29
[CI] Add dispatch_job script

boomanaiden154:
This patch adds the dispatch_job python script. This script is designed to be
invoked from within a llvm-zorg AnnotatedBuilder and spawn jobs inside of
kubernetes pods on the premerge cluster. This is not directly integrated
into an AnnotatedBuilder given the additional dependencies that we have
on the kubernetes client API.


premerge/buildbot/dispatch_job.py
```diff
@@ -0,0 +1,179 @@
...
+import sys
+import logging
+import time
+import dateutil
+import datetime
+import json
+
+import kubernetes
+
+PLATFORM_TO_NAMESPACE = {"Linux": "llvm-premerge-linux-builbot"}
```
cmtice: typo: "builbot" is missind the 'd'

premerge/buildbot/dispatch_job.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Args:
+      pod_name: The name of the pod to read logs from.
+      namespace: The namespace the pod is in.
+      v1_api: The kubernetes API instance to use for querying logs.
+
+    Returns:
+      A list of strings representing the log lines.
+    """
+    logs = v1_api.read_namespaced_pod_log(
```
cmtice, line 80: I really do not know: is "namespaced" a typo or is it correct?

premerge/buildbot/dispatch_job.py
boomanaiden154: Thanks, fixed.

I remember this being finnicky when testing locally with minikube. It turns out I had the namespace spelled incorrectly and then copied+pasted it into here...

🤦

premerge/buildbot/dispatch_job.py
boomanaiden154: Namespaced is correct here.

https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#read_namespaced_pod_log

PR #528, opened 2025-07-29
[CI] Add workflow to build premerge-buildbot container

boomanaiden154:
This patch sets up a workflow in a similar vein to all the other container
images for building the premerge-buildbot container and publishing it to
GHCR. There is a lot of duplicate code and eventually this should be
refactored into an action. This is on my TODO list.


.github/workflows/build-premerge-buildbot-container.yml
```diff
@@ -0,0 +1,76 @@
...
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/build-premerge-builtbot-container.yml
+      - 'premerge/builtbot/**'
```
cmtice: in line 12, is 'builtbot' a typo? (I'm also wondering about the use of 'builtbot' in the yml file names on lines 11 & 15...)

boomanaiden154: Yep. I was wondering why the job wasn't actually running...

Thank you for catching that.

PR #538, opened 2025-08-05
[CI] Setup Buildbot Infra for Premerge Builders

boomanaiden154:
This patch sets up the buildbot infrastructure for the premerge builders.


buildbot/osuosl/master/config/builders.py
```diff
@@ -3535,7 +3535,25 @@
                     checkout_llvm_sources=False,
                     script_interpreter=None,
                     clean=True)},
-
+    
+    # Builders that test the premerge configuration
```
cmtice, line 3539: *I* know that these builders are for post-submit testing of the premerge configuration, but nothing in the names or comments here makes that at all clear.  Ideally something in the builder names themselves would make this clear. Otherwise, at the very least there should be a very good set of comments here making that clear.  We don't want people confusing this set up with the actual premerge testing infrastructure.

buildbot/osuosl/master/config/workers.py
```diff
@@ -403,6 +403,12 @@ def get_all():
         create_worker("rise-worker-3", properties={'jobs' : 32}, max_builds=1),
         create_worker("rise-worker-4", properties={'jobs' : 32}, max_builds=1),
 
+        # Builders that run the premerge configuration
```
cmtice, line 406: Same comment as above (about names & comments & being very clear that this is NOT the actual premerge testing infra).

boomanaiden154: It would be difficult to change the worker names at this point given they're already set up on the staging buildbot. I've added a comment here better contextualizing where things are setup.

buildbot/osuosl/master/config/builders.py
boomanaiden154: Added a comment. The bots that previously tested the premerge configuration were also named this. I think keeping it consistent makes sense.

There shouldn't be too much room for confusion with the names given you can only see results for commits landed in `main` and buildbot only supports testing changes postcommit.

cmtice: Ok. Thanks for the comments. :-)

buildbot/osuosl/master/config/workers.py
cmtice: OK, I understand. Thanks for adding the comments.

buildbot/osuosl/master/config/builders.py
```diff
@@ -3535,7 +3535,29 @@
...
+    # These builders are specifically for running the premerge configuration
+    # postcommit (after changes have landed in main). The configuration for
+    # running these checks premerge exists in the monorepo inside the
+    # .github/workflows/premerge.yaml file.
+    {'name': "premerge-monolithic-linux",
+     'workernames': ["premerge-us-central-linux", "premerge-us-west-linux"],
+     'builddir': "premerge-monolithic-linux",
+     'factory': AnnotatedBuilder.getAnnotatedBuildFactory(
+                    script="premerge/dispatch_job.py",
```
cmtice, line 3548: Hasn't dispatch_job.py moved to a different directory? (Same comment for line 3557 below).

buildbot/osuosl/master/config/builders.py
```diff
@@ -3535,7 +3535,29 @@
...
+                    checkout_llvm_sources=False,
+                    extra_args=["Linux"],
+                    depends_on_projects=["bolt", "clang", "clang-tools-extra", "compiler-rt", "flang", "flang-rt", "libc", "libclc", "lld", "llvm", "mlir", "polly"])},
```
cmtice: Is there any reason we're not including 'lldb' in the projects list? (Same question for line 3560).

buildbot/osuosl/master/config/builders.py
boomanaiden154: It's in the `premerge` subdirectory of the new folder. I moved it to that folder since the `script` path here is relative to the root of that directory.

buildbot/osuosl/master/config/builders.py
boomanaiden154: It should be in there. I've added it. I copied and pasted them from the previous builders and it looks like I wasn't careful enough...

buildbot/osuosl/master/config/builders.py
```diff
@@ -3535,7 +3535,29 @@
...
+                    depends_on_projects=["bolt", "clang", "clang-tools-extra", "compiler-rt", "flang", "flang-rt", "libc", "libclc", "lld", "lldb", "llvm", "mlir", "polly"])},
+    
+    {'name': "premerge-monolithic-windows",
+     'workernames': ["premerge-us-central-windows", "premerge-us-west-windows"],
+     'builddir': "premerge-monolithic-windows",
+     'factory': AnnotatedBuilder.getAnnotatedBuildFactory(
+                    script="premerge/dispatch_job.py",
+                    checkout_llvm_sources=False,
+                    extra_args=["Windows"],
+                    depends_on_projects=["clang-tools-extra", "clang", "libclc", "lld", "llvm", "mlir", "polly"])},
```
cmtice, line 3560: I think you  need to add 'lldb' here too?

boomanaiden154: We don't test lldb on Windows builds currently. https://github.com/llvm/llvm-project/issues/132800

cmtice: Ah ok. :-)

cmtice (approved):

PR #542, opened 2025-08-05
[CI] Add Terraform Plumbing for Postcommit CI

boomanaiden154:
This patch does all the Terraform plumbing necessary for setting up the
postcommit buildbots. This leaves out the actual deployments for a future
patch.


premerge/gke_cluster/main.tf
```diff
@@ -78,6 +78,37 @@ resource "google_container_node_pool" "llvm_premerge_linux" {
   }
 }
 
+resource "google_container_node_pool" "llvm_buildbot_linux" {
```
cmtice, line 84: In my opinion, the names should mention postsubmit, to make it very clear what these are for.

premerge/gke_cluster/main.tf
```diff
@@ -78,6 +78,37 @@ resource "google_container_node_pool" "llvm_premerge_linux" {
...
+  name               = "llvm-buildbot-linux"
+  location           = var.region
+  cluster            = google_container_cluster.llvm_premerge.name
+  initial_node_count = 0
+
+  autoscaling {
+    total_min_node_count = 0
+    total_max_node_count = 3
```
cmtice, line 92: Is 3 really big enough for the max_node_count?

premerge/gke_cluster/main.tf
```diff
@@ -147,6 +178,53 @@ resource "google_container_node_pool" "llvm_premerge_windows_2022" {
   }
 }
 
+resource "google_container_node_pool" "llvm_buildbot_window_2022" {
```
cmtice, line 187: Same comment about the names.

premerge/gke_cluster/main.tf
```diff
@@ -147,6 +178,53 @@ resource "google_container_node_pool" "llvm_premerge_windows_2022" {
...
+  name               = "llvm-buildbot-windows-2022"
+  location           = var.region
+  cluster            = google_container_cluster.llvm_premerge.name
+  initial_node_count = 0
+
+  autoscaling {
+    total_min_node_count = 0
+    total_max_node_count = 3
```
cmtice, line 195: Same question about the max_node_count.

premerge/main.tf
```diff
@@ -121,6 +121,22 @@ data "google_secret_manager_secret_version" "grafana_token" {
   secret = "llvm-premerge-testing-grafana-token"
 }
 
+data "google_secret_manager_secret_version" "us_central_linux_buildbot_password" {
```
cmtice, line 127: Same comment about names.

cmtice: For entire file.

premerge/main.tf
```diff
@@ -121,6 +121,22 @@ data "google_secret_manager_secret_version" "grafana_token" {
...
+  secret = "llvm-buildbot-linux-us-central"
+}
+
+data "google_secret_manager_secret_version" "us_central_windows_buildbot_password" {
+  secret = "llvm-buidbot-windows-us-central"
```
cmtice: typo: "buidbot" is missing the 'l'

premerge/premerge_resources/main.tf
```diff
@@ -47,6 +47,48 @@ resource "kubernetes_namespace" "llvm_premerge_windows_2022_runners" {
   }
 }
 
+resource "kubernetes_namespace" "llvm_premerge_linux_buildbot" {
```
cmtice, line 53: Same comment about names.

cmtice: Next file,too.

premerge/main.tf
boomanaiden154: Fixed, thanks for catching that.

premerge/gke_cluster/main.tf
boomanaiden154: Discussion above.

premerge/gke_cluster/main.tf
boomanaiden154: We get three nodes per cluster, for a total of six. We only have 512 cores, so we can only do a total of eight per platform across both clusters. I think this should be enough (we managed with four with pretty much no queueing before).

It's easy to add more/change machine shape if we need to at this point too.

premerge/gke_cluster/main.tf
boomanaiden154: Having `buildbot` in all the names was supposed to achieve that purpose.

I feel that `postcommit_buildbot` is a bit redundant given buildbot can only be used for postcommit testing. If you feel differently I can change up the naming though.

cmtice: To me 'buildbot' just meant a machine/configuration that builds & runs tests. pre vs post never came into it.  If you don't want to change the names, how about adding a comment pointing out that all the 'buildbot' things refer to postcommit building/testing?

premerge/gke_cluster/main.tf
cmtice: ok. Thanks for the explanation.

premerge/main.tf
cmtice: Again, a comment clarifying what the 'buildbot'-named things are would be appreciated.

premerge/premerge_resources/main.tf
cmtice: Copy the buildbot-named things comment here too?

premerge/premerge_resources/variables.tf
```diff
@@ -104,3 +104,23 @@ variable "windows_2022_object_cache_gcp_service_account_email" {
   description = "The email associated with the service account for accessing the object cache on Windows."
   type        = string
 }
+
+variable "linux_buildbot_name" {
```
cmtice, line 108: In the descriptions you can say something about being for post commit testin.

premerge/gke_cluster/main.tf
boomanaiden154: Added a comment.

premerge/main.tf
boomanaiden154: Added a comment on top of all the secrets.

premerge/premerge_resources/main.tf
boomanaiden154: Done.

premerge/gke_cluster/main.tf
cmtice: Thanks!

cmtice (approved):

PR #543, opened 2025-08-05
Make some edits to the libc windows build config

amykhuang:
Update the configs for a windows build.


amykhuang:
Oh ok, I'll just remove that for now then.

PR #544, opened 2025-08-06
[CI] Add Buildbot Deployments

boomanaiden154:
This patch adds the buildbot deployments. These are pods designed to run the
buildbot worker in the service pool and then spawn jobs onto a node pool
inside their own pods.


premerge/premerge_resources/main.tf
```diff
@@ -77,7 +77,7 @@ resource "kubernetes_secret" "linux_buildbot_password" {
 resource "kubernetes_secret" "windows_buildbot_password" {
   metadata {
     name      = "windows-buildbot-password"
-    namespace = kubernetes_namespace.llvm_premerge_windows_2022_buildbot.name
+    namespace = "llvm-premerge-windows-buildbot"
```
cmtice: Not windows 2022?

cmtice (approved):

PR #553, opened 2025-08-15
[CI] Terraform Plubming for GCS Caching in Buildbot

boomanaiden154:
Add all of the service accounts/IAM bindings needed for GCS caching in the buildbot namespaces. Use separate service accounts in case we want to restrict the permissions between the PR checks and postcommit buildbots differently.


boomanaiden154:
Clicked the wrong buttons on Github, this isn't the one from spr...

PR #554, opened 2025-08-15
[CI] Update dispatch_job.py to setup sccache

boomanaiden154:
Now that we have buckets available and plumbed through terraform, setup
dispatch_job.py to actually setup sccache so the build scripts will use
the remote cache.

### DIFF
--- a/premerge/premerge_resources/variables.tf
+++ b/premerge/premerge_resources/variables.tf
@@ -61,7 +61,7 @@ variable "runner_group_name" {
 
 variable "libcxx_runner_image" {
   type    = string
-  default = "ghcr.io/llvm/libcxx-linux-builder:16f046281bf1a11d344eac1bc44d11f3e50e3b5d"
+  default = "ghcr.io/llvm/libcxx-linux-builder:36d31b0c008b2716329b5c9990f583decf919819"
 }
 
 variable "libcxx_release_runner_image" {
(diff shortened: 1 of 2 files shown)


zorg/buildbot/builders/annotated/premerge/dispatch_job.py
```diff
@@ -94,6 +99,10 @@ def start_build_linux(commit_sha: str, k8s_client) -> str:
         "export CC=clang",
         "export CXX=clang++",
         "export POSTCOMMIT_CI=1",
+        f"export SCCACHE_GCS_BUCKET={bucket_name}",
```
cmtice, line 102: Don't you need a '$' in front of the '{' ?

zorg/buildbot/builders/annotated/premerge/dispatch_job.py
```diff
@@ -104,14 +113,18 @@ def start_build_linux(commit_sha: str, k8s_client) -> str:
...
-def start_build_windows(commit_sha: str, k8s_client):
+def start_build_windows(commit_sha: str, bucket_name: str, k8s_client):
     """Starts a pod to build/test on Windows at the specified SHA."""
     pod_name = f"build-{commit_sha}"
     bash_commands = [
         "git clone --depth 100 https://github.com/llvm/llvm-project",
         "cd llvm-project",
         f"git checkout {commit_sha}",
         "export POSTCOMMIT_CI=1",
+        f"export SCCACHE_GCS_BUCKET={bucket_name}",
```
cmtice, line 124: Same question (about missing '$').

zorg/buildbot/builders/annotated/premerge/dispatch_job.py
boomanaiden154: Python fstrings don't need a `$` in front of the substitution. It would just appear as a $ in the resulting string

zorg/buildbot/builders/annotated/premerge/dispatch_job.py
boomanaiden154: Same as above.

PR #556, opened 2025-08-19
Also notify @cmtice for lldb-x86_64-debian failures

labath:


buildbot/osuosl/master/config/status.py
```diff
@@ -242,7 +242,7 @@ def getReporters():
         reporters.MailNotifier(
             fromaddr = status_email_fromaddr,
             sendToInterestedUsers = False,
-            extraRecipients = ["labath@google.com"],
+            extraRecipients = ["labath@google.com", "cmtice@google.com"],
```
labath, line 245: Can also be `cmtice+something` for easier filtering.

PR #559, opened 2025-08-19
[CI] Add more premerge buildbot workers

boomanaiden154:
The original intention for the premerge workers was that they could handle multiple builds at the same time. However, buildbot does not support a worker running more than one job at a time for a single builder, which is how we have things setup. Switch to having three builders per platform per cluster rather than one builder with max_builders set to 3.


gkistanova (approved):
LGTM.

PR #561, opened 2025-08-20
Add profcheck init script

boomanaiden154:
This is to be run on VM startup to setup the machine and start the buildbot.

### DIFF
--- a/buildbot/google/scripts/profcheck.sh
+++ b/buildbot/google/scripts/profcheck.sh
@@ -30,6 +30,12 @@ sudo -u buildbot buildbot-worker create-worker -f --allow-shutdown=signal $BOT_D
 
 {
   echo "To reproduce locally, use a standard CMake invocation with -DLLVM_ENABLE_PROFCHECK=ON and -DLLVM_LIT_ARGS='--exclude-xfail'"
+  echo "Example:"
+  echo "cmake -GNinja"
+  echo "  -DCMAKE_BUILD_TYPE=Release"
+  echo "  -DLLVM_ENABLE_ASSERTIONS=ON"
+  echo "  -DLLVM_LIT_ARGS='--exclude-xfail'"
+  echo "  -DLLVM_ENABLE_PROFCHECK=ON"
   echo
   uname -a | head -n1
   date
--- a/buildbot/google/scripts/profcheck.sh
+++ b/buildbot/google/scripts/profcheck.sh
@@ -13,7 +13,7 @@ rm -rf /b
 BOT_DIR=/b
 SERVER_PORT=9994
 WORKER_NAME="$(hostname)"
-WORKER_PASSWORD="7DcR2b^M.%uweylV9m"
+WORKER_PASSWORD="$(gsutil cat gs://sanitizer-buildbot/$(hostname)-password)"
 
 userdel buildbot | true
 groupadd -f buildbot
@@ -29,7 +29,7 @@ sudo -u buildbot buildbot-worker create-worker -f --allow-shutdown=signal $BOT_D
 } > $BOT_DIR/info/admin
 
 {
-  echo "How to reproduce locally: https://github.com/google/ml-compiler-opt/wiki/BuildBotReproduceLocally"
+  echo "To reproduce locally, use a standard CMake invocation with -DLLVM_ENABLE_PROFCHECK=ON and -DLLVM_LIT_ARGS='--exclude-xfail'"
   echo
   uname -a | head -n1
   date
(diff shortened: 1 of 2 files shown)


buildbot/google/scripts/profcheck.sh
```diff
@@ -0,0 +1,42 @@
+set -ex
+
+apt-get update
+apt-get install -y python3 python3-pip cmake ninja-build git ccache lsb-release wget software-properties-common gnupg wget
+pip3 install --break-system-packages buildbot-worker==3.11.7
+
+bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)" 20
+ln -sf /usr/bin/clang-20 /usr/bin/cc
```
mtrofin, line 8: shouldn't we use `update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-20 100` and similar?

buildbot/google/scripts/profcheck.sh
```diff
@@ -0,0 +1,42 @@
...
+ln -sf /usr/bin/clang++-20 /usr/bin/c++
+ln -sf /usr/bin/ld.lld-20 /usr/bin/ld
+
+rm -rf /b
+BOT_DIR=/b
+SERVER_PORT=9994
+WORKER_NAME="$(hostname)"
+WORKER_PASSWORD="7DcR2b^M.%uweylV9m"
```
mtrofin: umm... no?

create or reuse a bucket and use `gsutil`.

buildbot/google/scripts/profcheck.sh
```diff
@@ -0,0 +1,42 @@
...
+sudo -u buildbot buildbot-worker create-worker -f --allow-shutdown=signal $BOT_DIR lab.llvm.org:$SERVER_PORT \
+   "${WORKER_NAME}" "${WORKER_PASSWORD}"
+
+{
+  echo "Mircea Trofin <mtrofin@google.com>"
+  echo "Aiden Grossman <aidengrossman@google.com>"
+} > $BOT_DIR/info/admin
+
+{
+  echo "How to reproduce locally: https://github.com/google/ml-compiler-opt/wiki/BuildBotReproduceLocally"
```
mtrofin: nope :)

buildbot/google/scripts/profcheck.sh
boomanaiden154: 🤦 

Testing that I meant to switch over and didn't. I've contacted Galina to roll the token and changed this to use the standard bucket setup.

buildbot/google/scripts/profcheck.sh
boomanaiden154: Fixed. I've left it at just adding CMake options for now since the setup is way less complicated than the ml-compiler-opt builds. I think we can adjust if there turn out to be weird effects between the buildbot and standard configurations.

buildbot/google/scripts/profcheck.sh
boomanaiden154: What's the benefit of doing that?

We also want CMake to pick it up and by default it uses whatever is in `/usr/bin/cc` and `/usr/bin/c++`.

buildbot/google/scripts/profcheck.sh
```diff
@@ -29,7 +29,7 @@ sudo -u buildbot buildbot-worker create-worker -f --allow-shutdown=signal $BOT_D
 } > $BOT_DIR/info/admin
 
 {
-  echo "How to reproduce locally: https://github.com/google/ml-compiler-opt/wiki/BuildBotReproduceLocally"
+  echo "To reproduce locally, use a standard CMake invocation with -DLLVM_ENABLE_PROFCHECK=ON and -DLLVM_LIT_ARGS='--exclude-xfail'"
```
mtrofin, line 32: nit: I'd add an example of one that we know works (e.g. the one we use), just in case there are quirky cmake configs that one may think of as "standard".

buildbot/google/scripts/profcheck.sh
mtrofin: cmake would - afaik it's the more canonical way of ultimately creating those links. it's still 3 lines. Up to you.

boomanaiden154: It won't pick it up if we don't explicitly set `-DCMAKE_C_COMPILER` or `-DCMAKE_CXX_COMPILER`. That is part of the CMake config which requires a buildbot restart which I would prefer to avoid if possible.

Given it's about the same complexity, I'll leave it as is.

buildbot/google/scripts/profcheck.sh
boomanaiden154: Good point. Added.

PR #570, opened 2025-08-22
Switch sanitizer-windows bot over to runtimes build

boomanaiden154:
This enables us to catch issues like
https://g-issues.chromium.org/issues/437182411#comment20 at postcommit testing time rather than when trying to integrate into downstream projects.


rnk (approved):

vitalybuka (approved):

rnk:
Recent errors on the bot suggest that it is configured to do a 32-bit build:
```

[3895/4440] Linking CXX executable bin\bugpoint.exe
FAILED: bin/bugpoint.exe 
cmd.exe /C "cd . && "C:\Program Files\CMake\bin\cmake.exe" -E vs_link_exe --..." failed (exit code 1318) with the following output:
LINK : the 32-bit linker (C:\PROGRA~2\MIB055~1\2019\PROFES~1\VC\Tools\MSVC\1429~1.301\bin\Hostx86\x86\link.exe) ran out of heap space and is going to restart linking with a 64-bit linker
LINK : fatal error LNK1318: Unexpected PDB error; RPC (23) '(0x000006BA)'
```

Let's not do that, let's make sure it's an x64 build.

PR #575, opened 2025-08-26
Fix ClangBuilder for running LNT on Windows

omjavaid:
This PR adds support for running the LLVM test-suite on Windows buildbots using ClangBuilder. The changes ensure that the correct executable paths and file extensions are used when running Python and LNT scripts on Windows.

This depends on llvm/llvm-project#155226 which adds llvm-lit.cmd wrapper to make llvm-lit directory runnable on windows.

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -533,17 +533,18 @@ def _getClangCMakeBuildFactory(
                                    env=env))
 
         # Get generated python, lnt
-        python = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/python')
-        lnt = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/lnt')
-        lnt_setup = util.Interpolate('%(prop:builddir)s/test/lnt/setup.py')
+        python = InterpolateToPosixPath('%(prop:builddir)s/test/sandbox/Scripts/python')
+        lnt_ext = '.exe' if vs else ''
+        lnt = InterpolateToPosixPath(f'%(prop:builddir)s/test/sandbox/Scripts/lnt{lnt_ext}')
+        lnt_setup = InterpolateToPosixPath('%(prop:builddir)s/test/lnt/setup.py')
 
         # Paths
-        sandbox = util.Interpolate('%(prop:builddir)s/test/sandbox')
-        test_suite_dir = util.Interpolate('%(prop:builddir)s/test/test-suite')
+        sandbox = InterpolateToPosixPath('%(prop:builddir)s/test/sandbox')
+        test_suite_dir = InterpolateToPosixPath('%(prop:builddir)s/test/test-suite')
 
         # Get latest built Clang (stage1 or stage2)
-        cc = util.Interpolate(f'%(prop:builddir)s/{compiler_path}/bin/{cc}')
-        cxx = util.Interpolate(f'%(prop:builddir)s/{compiler_path}/bin/{cxx}')
+        cc = InterpolateToPosixPath(f'%(prop:builddir)s/{compiler_path}/bin/{cc}')
+        cxx = InterpolateToPosixPath(f'%(prop:builddir)s/{compiler_path}/bin/{cxx}')
 
         # LNT Command line (don't pass -jN. Users need to pass both --threads
         # and --build-threads in nt_flags/test_suite_flags to get the same effect)
@@ -558,7 +559,7 @@ def _getClangCMakeBuildFactory(
             # Append any option provided by the user
             test_suite_cmd.extend(nt_flags)
         else:
-            lit = util.Interpolate(f'%(prop:builddir)s/{stage1_build}/bin/llvm-lit')
+            lit = InterpolateToPosixPath(f'%(prop:builddir)s/{stage1_build}/bin/llvm-lit')
             test_suite_cmd = [python, lnt, 'runtest', 'test-suite',
                               '--no-timestamp',
                               '--sandbox', sandbox,
@@ -567,7 +568,7 @@ def _getClangCMakeBuildFactory(
                               '--cxx', cxx,
                               '--use-lit', lit,
                               # Carry on building even if there is a failure.
-                              '--build-tool-options', '"-k"']
+                              '--build-tool-options', '"-k 0"' if '--use-make=ninja' in testsuite_flags else '"-k"']
             # Enable fortran if flang is checked out
             if checkout_flang:
                 fortran_flags = [


zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -533,17 +533,18 @@ def _getClangCMakeBuildFactory(
                                    env=env))
 
         # Get generated python, lnt
-        python = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/python')
-        lnt = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/lnt')
-        lnt_setup = util.Interpolate('%(prop:builddir)s/test/lnt/setup.py')
+        python = InterpolateToPosixPath('%(prop:builddir)s/test/sandbox/Scripts/python')
```
DavidSpickett, line 536: Thouht this was wrong, but despite the name, InterpolateToPosixPath does replace the posix separator with the windows one when on Windows.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -567,7 +568,7 @@ def _getClangCMakeBuildFactory(
                               '--cxx', cxx,
                               '--use-lit', lit,
                               # Carry on building even if there is a failure.
-                              '--build-tool-options', '"-k"']
+                              '--build-tool-options', '"-k 0"' if '--use-make=ninja' in testsuite_flags else '"-k"']
```
DavidSpickett, line 571: I take from this that on Windows, you are going to pass `--use-make=ninja`?

Since make -k has no argument but ninja requires some argument, which is annoying but it is what it is.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -533,17 +533,18 @@ def _getClangCMakeBuildFactory(
                                    env=env))
 
         # Get generated python, lnt
-        python = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/python')
-        lnt = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/lnt')
-        lnt_setup = util.Interpolate('%(prop:builddir)s/test/lnt/setup.py')
+        python = InterpolateToPosixPath('%(prop:builddir)s/test/sandbox/Scripts/python')
+        lnt_ext = '.exe' if vs else ''
+        lnt = InterpolateToPosixPath(f'%(prop:builddir)s/test/sandbox/Scripts/lnt{lnt_ext}')
```
DavidSpickett, line 538: Surprised that you need to specify .exe, expected Windows would find lnt.exe from just lnt, but I presume you tested this.

zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -533,17 +533,18 @@ def _getClangCMakeBuildFactory(
                                    env=env))
 
         # Get generated python, lnt
-        python = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/python')
-        lnt = util.Interpolate('%(prop:builddir)s/test/sandbox/bin/lnt')
-        lnt_setup = util.Interpolate('%(prop:builddir)s/test/lnt/setup.py')
+        python = InterpolateToPosixPath('%(prop:builddir)s/test/sandbox/Scripts/python')
+        lnt_ext = '.exe' if vs else ''
```
DavidSpickett, line 537: `vs` True does always mean Windows, but `vs` False does not mean the host is not Windows. In reality there are no non-vs Windows builds because we always need at least some part of the Microsoft toolchain, right?

DavidSpickett: Which makes `vs` good enough for "is the host Windows" checks.

zorg/buildbot/builders/ClangBuilder.py
omjavaid: Yes ninja -k required 0 as argument while for make -k is enough.

zorg/buildbot/builders/ClangBuilder.py
omjavaid: Yes this works well on windows. We fixed some of the paths in ClangBUilder for the normal WoA bots but as we werent running the testsuite some of the paths left unchanged which are fixed in this patch.

zorg/buildbot/builders/ClangBuilder.py
omjavaid: lnt.exe is an executable shim for lnt.py i guess. so we run python lnt.exe and it gets launched properly but if we python lnt it doenst know which one to choose.

zorg/buildbot/builders/ClangBuilder.py
omjavaid: vs=None means non Windows MSVC builder.
vs=True mean MSVC builder and it should run vcvarsall.bat as well.
vs=manual means buildbot worker is started in an environment after running vcvarsall.bat or similar.
when vs is set we also set cc to clang.exe instead of clang. Similar cxx, and flang get .exe appended.

DavidSpickett: Ok so while you could probably run something on Windows that set `vs` to false, clearly no one does because it would have failed already to find flang/clang/etc.

zorg/buildbot/builders/ClangBuilder.py
DavidSpickett, line 536: This is not correct for Linux:
```
/home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox/Scripts/python -m pip install setuptools
 in dir /home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox (timeout 1200 secs)
 watching logfiles {}
 argv: [b'/home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox/Scripts/python', b'-m', b'pip', b'install', b'setuptools']
 environment:
  HOME=/home/tcwg-buildbot
  HOSTNAME=linaro-clang-aarch64-lld-2stage
  LANG=en_US.UTF-8
  LOGNAME=tcwg-buildbot
  MAIL=/var/mail/tcwg-buildbot
  NO_STOP_MESSAGE=1
  PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
  PWD=/home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox
  SHELL=/bin/bash
  SHLVL=0
  SUDO_COMMAND=/bin/bash -c buildbot-worker restart --nodaemon /home/tcwg-buildbot/worker
  SUDO_GID=0
  SUDO_UID=0
  SUDO_USER=root
  TERM=xterm
  USER=tcwg-buildbot
  _=/usr/bin/buildbot-worker
 using PTY: False
Upon execvpe b'/home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox/Scripts/python' [b'/home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox/Scripts/python', b'-m', b'pip', b'install', b'setuptools'] in environment id 251397130647040
:Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/twisted/internet/process.py", line 397, in _fork
    self._execChild(path, uid, gid, executable, args, environment)
  File "/usr/lib/python3/dist-packages/twisted/internet/process.py", line 468, in _execChild
    os.execvpe(executable, args, environment)
  File "/usr/lib/python3.10/os.py", line 584, in execvpe
    _execvpe(file, args, env)
  File "/usr/lib/python3.10/os.py", line 598, in _execvpe
    exec_func(file, *argrest)
FileNotFoundError: [Errno 2] No such file or directory: b'/home/tcwg-buildbot/worker/clang-aarch64-lld-2stage/test/sandbox/Scripts/python'
program finished with exit code 1
elapsedTime=0.048819
```
Needs to be `/bin`:
```
$ ./worker/clang-aarch64-lld-2stage/test/sandbox/bin/python
Python 3.10.12 (main, May 27 2025, 17:12:29) [GCC 11.4.0] on linux
```

PR #576, opened 2025-08-26
Add flang-arm64-windows-msvc-testsuite builder

omjavaid:
This PR adds a new buildbot to run LLVM testsuite on windows. It adds a new builder and worker pair. The testsuite will run after the first stage build skipping the ninja check step as we cover that on other buildbots. This will primarily be the builder to test that ensures LNT and LLVM testsuite are running without regressions on windows. We will later build on it to add support for test more as our Windows hardware availability improves.

This depends on #575 and llvm/llvm-project#155226.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2631,6 +2631,29 @@ all += [
                         "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
                         "-DLLVM_CCACHE_BUILD=ON"])},
 
+    {'name' : "flang-arm64-windows-msvc-testsuite",
+    'tags'  : ["flang"],
+    'workernames' : ["linaro-armv8-windows-msvc-06"],
+    'builddir': "flang-arm64-win-msvc-ts",
+    'factory' : ClangBuilder.getClangCMakeBuildFactory(
+                    vs="manual",
+                    clean=False,
+                    checkout_flang=True,
+                    checkout_lld=True,
+                    runTestSuite=True,
+                    testStage1=False,
+                    testsuite_flags=[
+                        '--cmake-define', "TEST_SUITE_SUBDIRS='Fortran'",
+                        '--use-make=ninja',
+                        '--threads=8',
+                        '--build-threads=8'],
+                    extra_cmake_args=[
+                        "-DLLVM_TARGETS_TO_BUILD=AArch64",
+                        "-DCLANG_DEFAULT_LINKER=lld",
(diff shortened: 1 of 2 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -2651,6 +2651,29 @@
                         "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
                         "-DLLVM_CCACHE_BUILD=ON"])},
 
+    {'name' : "flang-arm64-windows-msvc-testsuite",
+    'tags'  : ["flang"],
+    'workernames' : ["linaro-armv8-windows-msvc-06"],
+    'builddir': "flang-arm64-win-msvc-ts",
+    'factory' : ClangBuilder.getClangCMakeBuildFactory(
+                    vs="manual",
```
DavidSpickett, line 2639: Manual means what exactly?

Is that where buildbot is started in an environment where the Visual Studio environment is already applied?
```
vcvarsall.bat
buildbot-agent
```
Like that? Instead of buildbot running vcvars for each build.

buildbot/osuosl/master/config/builders.py
```diff
@@ -2651,6 +2651,29 @@
...
+                    testsuite_flags=[
+                        '--cmake-define', "TEST_SUITE_SUBDIRS='Fortran'",
+                        '--use-make=ninja',
+                        '--threads=8',
+                        '--build-threads=8'],
+                    extra_cmake_args=[
+                        "-DLLVM_TARGETS_TO_BUILD=AArch64",
+                        "-DCLANG_DEFAULT_LINKER=lld",
+                        "-DCMAKE_TRY_COMPILE_CONFIGURATION=Release",
+                        "-DCOMPILER_RT_BUILD_SANITIZERS=OFF",
```
DavidSpickett, line 2654: Where does the compiler-rt project get added? Do we just know that if you're going to run the test suite then you need the runtime libraries?

buildbot/osuosl/master/config/builders.py
omjavaid: manual means that its an MSVC bot but builder will not be responsible for setting up MSVC environment with vcvarsall.bat. I think when MSVC support was initially set up ClangBuilder was able to detect and run vcvarsall.bat
script but we never tested this on windows on arm.

buildbot/osuosl/master/config/builders.py
omjavaid: checkout_compiler_rt is True by default in ClangBuilder unless we specify it as False.

PR #578, opened 2025-08-27
[CI] Write Routine Maintenance Document

boomanaiden154:
This patch introduces a document that describes all of the routine maintenance actions that need to be performed on the premerge cluster, currently only version updates.

### DIFF
--- a/premerge/routine-maintenance.md
+++ b/premerge/routine-maintenance.md
@@ -51,7 +51,7 @@ be updated to the new version.
 3. The libc++ CI container - The `docker-compose` manifest at
 `libcxx/utils/ci/docker-compose.yml` needs to be updated by updating the
 `GITHUB_RUNNER_VERSION` variable to pull in the latest
-runner images using the [libc++ instructions](https://libcxx.llvm.org/Contributing.html#updating-the-ci-testing-container-images)
+runner binary using the [libc++ instructions](https://libcxx.llvm.org/Contributing.html#updating-the-ci-testing-container-images)
 
 ### Other Container Image Software
 


premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+
+1. The Linux CI container - The `Dockerfile` at
+`.github/workflows/containers/github-action-ci/Dockerfile` has an environment
+variable towards the bottom of the file called `GITHUB_RUNNER_VERSION` that
+needs to be updated to the new version.
+2. The Windows CI container - The `Dockerfile` at
+`.github/workflows/containers/github-action-ci-windows/Dockerfile` has an
+argument called `RUNNER_VERSION` near the bottom of the file that needs to
+be updated to the new version.
+3. The libc++ CI container - The `docker-compose` manifest at
```
cmtice, line 51: It's not clear what you mean by "The `docker-compose` manifest". Are you talking about the GITHUB_RUNNER_VERSION (in the docker-compose.yml file)? Or something else?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+`libcxx/utils/ci/docker-compose.yml` needs to be updated to pull in the latest
```
cmtice: Actually the entire #3 set of instructions is a bit confusing. The libc++ instructions you reference mention how to grab new container image that already contains a particular runner version, but they don't explicitly state how to create a container with the correct runner version? Which is what #1 and #2 seem to be about?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+runner images using the [libc++ instructions](https://libcxx.llvm.org/Contributing.html#updating-the-ci-testing-container-images)
+
+### Other Container Image Software
```
cmtice, line 56: Which repo alert would cause trigger this set of instructions?  (It's not clear that any of the repos mentioned above correspond to this).

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+
+The container images also contain many other pieces of software critical
+for building LLVM, like CMake, ninja, and the toolchain itself. Keeping
+most of these up to date is ideal.
+
+A large amount of the software comes from the distribution and thus does not
```
cmtice: "the distribution" -- What distribution? And how do you get notified that it's time to update it?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+need to be explicitly updated. We prefer to install software from the
+distribution when possible. However, this does mean that distribution
+updates are quite important. To update the distribution for the Linux container,
+perform the following steps:
+
+1. Modify `.github/workflows/containers/github-action-ci/Dockerfile` locally
+to use the latest `ubuntu:xx.04` image.
+2. Ensure that `monolithic-linux.sh` with all options enabled runs successfully.
```
cmtice, line 70: Are there instructions for how to test  monolithic-linux.sh with all options enabled?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+Push any needed changes.
+3. Push the updated `Dockerfile` and update the workflow in
+`.github/workflows/build-ci-container.yml` to use the correct image name.
```
cmtice: For #3: Can you be more explicit about what to update in build-ci-container.yml?  Do you mean the 'container_name" in the bulid-ci-container job? or something else? What's the right value to update it to?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+4. Update the runner configuration in zorg (`premerge/linux_runners_values.yaml`)
+to point to the new image.
+
+Updating explicitly versioned software (just LLVM in the Linux container, but
+most software in the Windows container) just requires bumping the version number
+and pushing the new image to the monorepo. In the Linux container, the LLVM
+version can be bumped by changing the `LLVM_VERSION` environment variable. In
```
cmtice, line 83: The "containers" you're talking about here refer to the Dockerfiles mentioned above? Under .github/workflows/containers/github-actions-ci{-windows}?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+the Windows container, versions are controlled by the `--version` flag passed
+to `choco install` commands.
```
cmtice: The Windows Dockerfile doesn't contain any 'choco install' commands for installing LLVM that I can see?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+[described steps](cluster-management.md#upgradingresetting-github-arc) to
+avoid any downtime.
+
+It is advised to do this during a portion of the day with light traffic as
+any upgrade will involve having an entire cluster down at times, which reduces
+capacity in half.
+
+### Windows Edition
+
+Whenever a new Windows Server datacenter edition (eg 2025) is supported by GKE,
```
cmtice, line 100: How do you get notifications about this?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+all of the windows infrastructure needs to be updated. This involves several
+steps that will take significant time:
+
+1. Modify `.github/workflows/containers/github-action-ci-windows/Dockerfile`
+locally to get it building on the new Windows Server version. This requires
+having a host at that version as Windows Server containers can only run on a
```
cmtice, line 106: "having a host at that version"?  A host windows machine?  Where/how do you get a local windows machine for this?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+host with the same edition. Changing the `FROM` line at the top of the
+`Dockerfile` has previously been enough.
+2. When that is ready, push your changes in a PR along with changes to
+`.github/workflows/build-ci-container-windows.yml` so that it uses the new
+Windows Server edition.
```
cmtice: So, this involved changing all the "windows-20xx" in that file (both the runs-on values and the container_name) to use the new "windows-20yy" version?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+3. Test locally that the pushed container image can run the
+`monolithic-windows.sh` script with all projects enabled successfully. Make
+and submit any changes that are needed.
```
cmtice: Roughly same question as above: How do you test monolithic-windows.sh with all projects enabled locally?

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+4. Duplicate all of the versioned Windows resources in the terraform
```
cmtice, line 128: So, for example,  if we're upgrading from windows-2022 to windows-2025, I would KEEP all the windows-2022 resources defined in the all the *.tf files (under premerge), but would now also have windows-2025 versions of them?

Also I assume I would  need to copy windows_2022_runner_values.yaml to windows_2025_runner_values.yaml, and update the values inside to use 2025 instead of 2022? 


Also, both architecture.md and cluster-management.md explicitly mention 'windows_2022' or 'windows-2022', so if the version gets updated then the docs need to be updated too.

cmtice: Also, should probably mention that this is in the llvm-zorg repo, since most of the other steps are in the llvm-project repo.

premerge/routine-maintenance.md
```diff
@@ -0,0 +1,116 @@
...
+configuration and apply it.
+5. Switch the workflow in `.github/workflows/premerge.yaml` to the new
+runner set.
+6. Sunset the existing sunset in 1-2 weeks to give time for people utilizing
```
cmtice: "Sunset the existing sunset" seems like a typo.  Also, you should be more explicit about what 'sunsetting' means here. I assume you mean: Delete all the old resources that you KEPT in step 4 when you duplicated everything.  Are there other steps beyond that?

boomanaiden154: s/sunset/runnerset

Also added some wording to clarify what sunsetting means (delete the old runnerset and any other associated resources).

premerge/routine-maintenance.md
boomanaiden154: Added some text to make this more clear.

premerge/routine-maintenance.md
boomanaiden154: Added some text. This assumes things go smoothly, but they should. The complicated bits should be in the Dockerfile.

premerge/routine-maintenance.md
boomanaiden154: Added some information on how to do this.

premerge/routine-maintenance.md
boomanaiden154: This should be done by creating a GCP VM. I mentioned that. The exact process is best described in the GCP docs which can be easily found by searching if someone isn't already familiar with the process.

premerge/routine-maintenance.md
boomanaiden154: I'm not aware of a good way. It's a pretty rare upgrade though, so subscribing to it might be more trouble than it is worth.

premerge/routine-maintenance.md
boomanaiden154: We don't install LLVM this way, but do install other packages this way. Added some wording.

premerge/routine-maintenance.md
boomanaiden154: Yes, that variable. Added wording.

premerge/routine-maintenance.md
boomanaiden154: None of the repositories listed above. These aren't necessarily that critical though as it is mostly a couple pieces of software on the Windows side like python and ninja that are pretty stable. You technically could subscribe to the releases there, but there might be some unnecessary churn if we keep exactly up to date.

premerge/routine-maintenance.md
boomanaiden154: The Linux distribution. Updates are rare (once every two years for Ubuntu LTS), so I'm not sure it's too worth having a notification for it.

premerge/routine-maintenance.md
boomanaiden154: Added how to find this.

premerge/routine-maintenance.md
boomanaiden154: Yes.

premerge/routine-maintenance.md
boomanaiden154: Yes. Added wording.

premerge/routine-maintenance.md
boomanaiden154: There's wording in the libc++ document to cover this exact case:

> To update the runner binary, bump the value of GITHUB_RUNNER_VERSION in libcxx/utils/ci/docker/docker-compose.yml.

s/images/binary though, which is probably the confusing part.

PR #582, opened 2025-08-29
Insert debugify-report into the test suite build directory

SLTozer:
As part of the DebugifyBuilder, we add cflag arguments to the test suite build step containing the path of a report file, where any bugs detected by debugify will be written. Debugify appends to the file rather than replacing it, as we wish to accumulate the bugs found in all invocations of Clang during the test suite build step to a single file. Currently however, this file is not cleaned up after builds, meaning that we are constantly appending bugs to the same file after each test run, rather than creating a fresh file - this contaminates the results for future builds, as bugs will never disappear from the list even if they no longer exist in LLVM. This patch moves the report file into the test suite build directory, which will always be cleaned before each test run, preventing the results of one run from seeping into subsequent runs.

### DIFF
--- a/zorg/buildbot/builders/DebugifyBuilder.py
+++ b/zorg/buildbot/builders/DebugifyBuilder.py
@@ -1,6 +1,7 @@
 from buildbot.plugins import util
 from buildbot.steps.shell import ShellCommand
 from zorg.buildbot.builders import TestSuiteBuilder
+from zorg.buildbot.builders.TestSuiteBuilder import test_suite_build_path
 from zorg.buildbot.commands.CmakeCommand import CmakeCommand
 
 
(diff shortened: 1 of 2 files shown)


DavidSpickett (approved):
I have no context for the builder overall, I would look at the existing builder but the site is down at the moment. This seems logical in isolation though.

Assuming that this is the way it works:
* test suite build dir is cleaned before each new run
* debugify report is written into test suite build dir
* a script is run on that report, which generates a further report, which is shown in the build results
* the original report file can now be discarded, and will be, when the next run happens

Then this LGTM.

I just had to check that you weren't saving the original report as an artifact because if so, you're missing changing the location to save it from.

PR #584, opened 2025-09-01
Drop unnecessary -treat-scalable-fixed-error-as-warning=false options

nikic:
This is already the default, and I'm dropping this option in https://github.com/llvm/llvm-project/pull/156336.


DavidSpickett (approved):
Assuming llvm reviewers agree with the changes there, this LGTM. We (Linaro) added this option on Arm's recommendation in the first place.

Land this first and once that's live on the buildbot, the llvm change can go in. To keep the bots working throughout.

PR #587, opened 2025-09-02
[CI] Add check-lit to postcommit testing

jayfoad:
`check-llvm` does not run tests for the `llvm-lit` utility, so run the `check-lit` target as well.


jayfoad:
I don't usually work on zorg. I don't know if this is the right place to implement this, and I don't know how to test it. Reviews welcome! Thanks.

boomanaiden154 (approved):
This only runs on the premerge postcommit builders, but that should be fine. We'll want to add coverage for this on the premerge side, but I can put up a patch for that.

PR #592, opened 2025-09-04
[ClangBuilder] Correct path of testsuite virtualenv's lnt

omjavaid:
This is a follow up to #586.

Fixes 60d2141a5073ce1330bc45e4cc827344af67894b.

We forgot to fix the lnt path to work on both windows and linux.

On Windows, it's Scripts/lnt, on Linux it's bin/lnt.


DavidSpickett (approved):
LGTM

PR #599, opened 2025-09-16
[CI] Add Design Doc for Premerge Advisor

boomanaiden154:
This patch adds a design document for the premerge advisor that we plan on implementing.

### DIFF
--- a/premerge/premerge-advisor.md
+++ b/premerge/premerge-advisor.md
@@ -53,9 +53,10 @@ explaining its findings.
 A key part of the premerge advisor is infrastructure to store and process build
 failure information so that it can be queried later. We plan on having jobs
 extract failure logs and upload them to a web server. This web server in
-addition to having an endpoint to accept uploads will have an endpoint that will
-accept test failure information (logs and filenames) and return whether or not
-they are broken at `main`, flaky, or novel test failures due to the PR.
+addition to having an endpoint to accept uploads of failure information without
+returning any analysis will have an explanation endpoint that will accept test
+failure information (logs and filenames) and return whether or not they are
+broken at `main`, flaky, or novel test failures due to the PR.
 
 For the premerge jobs running through Github Actions, we plan on using the
 existing `generate_test_report` scripts that are currently used for generating
@@ -63,17 +64,22 @@ summaries on job failures. When the job ends and there is a failure, there would
 be a script that runs, utilizing the `generate_test_report` library to extract
 failure information, and then uploads the information to the web server.
 Information on how the premerge jobs will query the server and display results
-about flaky/already broken tests is in
-[the section below](#informing-the-user). We plan on having both the premerge
-and [postcommit jobs](post-submit-testing.md) upload failure information to the
-web server. This enables the web server to know about failures that are not the
-result of mid-air collisions before postcommit testing has been completed.
-Postcommit testing can take upwards of 30 minutes, during which the premerge
-advisor would not be able to advise that these failures are due to a broken
-`main`. Data from both postcommit and premerge testing will be associated with a
-commit SHA for the base commit in main. Premerge testing data will additionally
-have a PR associated with it to enable disambiguating between failures occurring
-in `main` at a specific commit and failures only occurring in a PR.
+about flaky/already broken tests is in [the section below](#informing-the-user).
+We plan on having both the premerge and [postcommit
+jobs](post-submit-testing.md) upload failure information to the web server. This
+enables the web server to know about failures that are not the result of mid-air
+collisions before postcommit testing has been completed. Postcommit testing can
+take upwards of 30 minutes, during which the premerge advisor would not be able
+to advise that these failures are due to a broken `main`. Data from both
+postcommit and premerge testing will be associated with a commit SHA for the
+base commit in main. Premerge testing data will additionally have a PR
+associated with it to enable disambiguating between failures occurring in `main`
+at a specific commit and failures only occurring in a PR. All failure
+information will be stored in the database. When testing for a PR is completed
+and there are failures, premerge will then be able to query the advisor to
+explain any outstanding failures. The advisor will compare the logs from the
+premerge run to the database and return its results, namely whether or not any
+of the failures exist at `HEAD` or are known flakes.
 
 We plan on implementing the web server in python with the `flask` library. All
 contributors to the premerge infrastructure are already familiar with Python,
@@ -84,41 +90,54 @@ in terms of infrastructure setup, and is reliable.
 
 Given we have two identical clusters that need to be able to function
 independently of each other, we also need to duplicate the web server for the
-premerge advisor. Queries and failure information uploads will go to the
-cluster-local premerge advisor web server. Periodically (eg once every thirty
-seconds), the web server will query the web server on the other cluster to see
-if there is any new data that has not been propagated back to the other side yet.
-It is easy to figure out what one side is missing as Github workflow runs are
-numbered sequentially and git commits are also explicitly ordered. One side just
-needs to send the latest commit SHA and workflow run it has all previous data
-for, and the other side can reply with the data that it has past that point.
-Explicitly synchronizing everytime without assumptions about the state of the
-other side has benefits over just writing through, like ensuring that a cluster
-that has been down for a significant amount of time is seamlessly able to
-recover. Synchronization does not need to be perfect as test failures that are
-flaky or broken at head will almost certainly show up in both clusters
-relatively quickly, and minor discrepancies for queries between the clusters are
-not a big deal.
+premerge advisor. Queries will go to the cluster-local premerge advisor web
+server. Failure uploads will directly write to both premerge advisor containers.
+This could lead to situations where uploading to one server succeeds while it
+fails writing to the other one, resulting in inconsistent state between the
+clusters. This is not a major concern because in the rare event we do end up
+with inconsistent states, the only major concern would be inconsistent results
+returned between the clusters. Uploading directly to both premerge containers
+and explicitly allowing inconsistent state means we do not need to maintain any
+state replication/consistency machinery, which would add significant complexity
+and maintenance costs.
 
-### Identifying if Failures are Novel
+For the database, we will have two tables, one called `failures` to store
+failure information, and one called `commits` that will be used to map commit
+SHAs to SVN like indices. The `commits` table will simply have two columns,
+mapping SHAs to indices. The failure table is designed to have rows describing
+individual test failures, and the schema looks like the following:
+
+* `source_type` - Whether the failure information comes from a postcommit run or
+  from a premerge run.
+* `base_commit_sha` - The SHA of the base commit. For premerge runs, this is the
+  commit in `main` (or the user branch) that the PR is based on top of. For
+  postcommit runs, this is simply the SHA of the commit being tested.
+* `commit_index` - The index of the base commit, assuming the base commit is in
+  `main`.
+* `source_id` - The ID of the source. For buildbot runs, this is the buldbot run
+  number. For premerge runs, this is the Github Actions workflow run ID.
+* `test_file` - The test file/compilation unit where the failure occurred.
+* `failure_message` - Any failure message text associated with the failure.
+* `platform` - The platform (including OS and architecture) where the failure
+  occurred.
+
+### Classifying Test Failures
 
 When the web server is sent a request to explain a list of failures, it needs to
 be able to determine whether a failure has occurred previously. To do this, the
 web server will keep a list of currently active failures in `main` and a list of
 flaky tests. The list of flaky tests will be computed from historical data on
 startup from a rolling window of data. The exact heuristic is left to
-implementation as it will likely require some experimentation. The list of
-currently active failures will initially be taken from the last known postcommit
-run. If a new postcommit run shows additional failures that are not in the flake
-list, they will be added to the currently active failure list. Failures in the
-currently active list not present in the latest postcommit run will be removed
-from the currently active list. In addition to data from postcommit testing, if
-a PR lands with unexplained premerge failures, those failures are also added to
-the currently active failure list. If a PR based on a version of main with
-currently active failures no longer contains those failures, they will be
-removed from the list when that PR lands. We look at information from PRs to
-improve the latency of failures/passes making their way into the system with
-postcommit testing currently having a minimum latency of 30 minutes.
+implementation as it will likely require some experimentation, but will look at
+whether or not the test has been failing periodically over a long period of
+time. The list of currently active failures will initially be taken from the
+last known postcommit run. If a new postcommit run shows additional failures
+that are not in the flake list, they will be added to the currently active
+failure list. Failures in the currently active list not present in the latest
+postcommit run will be removed from the currently active list. In the future, we
+want to look at information from PRs to improve the latency of failures/passes
+making their way into the system due to postcommit testing currently having a
+minimum latency of 30 minutes.
 
 Failures will be identified as being the same through a combination of the
 test name and likely a fuzzy match of the test error output. The exact details
@@ -150,14 +169,12 @@ in `main` or as flaky.
 After the response from the server has been recieved, the workflow will then
 construct a comment. It will contain the failure information, and if relevant,
 information/links showing the tests are flaky (and the flakiness rate) or are
-broken in `main`. If all of the test failures are due to flakes or failures in
-`main`, the comment will indicate to the user that they can safely merge their
-PR despite the test failures. We plan to construct the comment in a manner
-similar to the code formatting action. We will create one comment on the first
-workflow failure and then update that comment everytime we get new data. This
-prevents creating much noise. This does mean that the comment might get buried
-in long running reviews, but those are not the common case and people should
-learn to expect to look for the comment earlier in the thread in such cases.
+broken in `main`. We plan to construct the comment in a manner similar to the
+code formatting action. We will create one comment on the first workflow failure
+and then update that comment everytime we get new data. This prevents creating
+too much noise. This does mean that the comment might get buried in long running
+reviews, but those are not the common case and people should learn to expect to
+look for the comment earlier in the thread in such cases.
 
 If all of the failures in the workflow were successfully explained by the
 premerge advisor as flaky or already broken in `main`, then the premerge
@@ -192,8 +209,7 @@ reduce the risk of "mid-air collisions", patches that pass premerge testing, but
 fail on `main` when landed due to the patch in its original state not being
 compatible with the new state of main. This would most likely involve merge
 queues which would introduce new CI load and are also not compatible with LLVM's
-existing practices for the same reason requiring premerge checks to be run
-before landing are not.
+existing practices.
 
 Doing an initial effort for deflaking tests is also not scalable from an
 engineering effort perspective. While we might be able to deflake existing
--- a/premerge/premerge-advisor.md
+++ b/premerge/premerge-advisor.md
@@ -203,13 +203,11 @@ scalable.
 
 In order to ensure that main is not broken by new patches, we need to ensure
 that every commit is tested directly on top of `main` before landing. This is
-not feasible given LLVM's current processes where pushing directly to main is a
-critical component of several developer's workflows. We would also need to
-reduce the risk of "mid-air collisions", patches that pass premerge testing, but
-fail on `main` when landed due to the patch in its original state not being
-compatible with the new state of main. This would most likely involve merge
-queues which would introduce new CI load and are also not compatible with LLVM's
-existing practices.
+not feasible given the presence of "mid-air collisions", patches that pass
+premerge testing, but fail on `main` when landed due to the patch in its
+original state not being compatible with the new state of main. This would most
+likely involve merge queues which would introduce new CI load and are also not
+compatible with LLVM's existing practices.
 
 Doing an initial effort for deflaking tests is also not scalable from an
 engineering effort perspective. While we might be able to deflake existing
--- a/premerge/premerge-advisor.md
+++ b/premerge/premerge-advisor.md
@@ -1,0 +1,168 @@
+# Premerge Advisor
+
+## Introduction
+
+While the premerge infrastructure is generally reliable, tests that are broken
+at HEAD, or tests that are flaky can significantly degrade the user experience.
+Failures unrelated to the PR being worked on can lead to warning fatigue which
+can compound this problem when people land failing tests into main that they
+believe are unrelated. The premerge advisor is designed to run after premerge
+testing and signal to the user whether the failures can be safely ignored because
+they are flaky or broken at head, or whether they should be investigated further.
+
+## Background/Motivation
+
+The usefulness of the premerge system is significantly impacted by how much
+people trust it. People trust the system less when the issues reported to them
+are unrelated to the changes that they are currently working on. False positives
+occur regularly whether due to flaky tests, or due to main being broken. Efforts
+to fix these issues at the source and keep main always green and deflake tests
+are laudable, but not a scalable solution to the problem of false positive
+failures. It is also not reasonable to expect PR authors to spend time
+familiarizing themselves with all known flaky tests and dig through postcommit
+testing logs to see if the failures in their premerge run also occur in main.
+These alternatives are further explored in the section on
+[alternatives considered](#alternatives-considered). Having tooling to
+automatically run through the steps that one would otherwise need to perform
+manually would ensure the analysis on every failed premerge run is thorough and
+likely to be correct.
+
+## Design
+
+The premerge advisor will consist of three main parts: jobs uploading failure
+information, a web server and database to store and query failure information,
+and tooling to write out a verdict about the failures to comments on Github.
+When a job runs premerge or postcommit and there are build/test failures, it
+will upload information to the web server containing information on the failure
+like the test/build action that failed and the exact log. The web server will
+then store this information in a format that makes it easy to query for later.
+Every premerge run that encounters build/test failures will then query the web
+server to see if there are any matching build/test failures for the version of
+`main` that the PR has been merged into. If there are, the web server can
+respond with the failure information and signal that the failures can be safely
+ignored for that premerge run. If the failures are novel, then the web server
+can signal that the failures should be investigated more thoroughly. If there
+are failures, the premerge advisor can then write out a comment on the PR
+explaining its findings.
+
+### Processing and Storing Failures
+
+A key part of the premerge advisor is infrastructure to store and process build
+failure information so that it can be queried later. We plan on having jobs
+extract failure logs and upload them to a web server. This web server in
+addition to having an endpoint to accept uploads will have an endpoint that will
+accept test failure information (logs and filenames) and return whether or not
+they are broken at `main`, flaky, or novel test failures due to the PR.
+
+For the premerge jobs running through Github Actions, we plan on using the
+existing `generate_test_report` scripts that are currently used for generating
+summaries on job failures. When the job ends and there is a failure, there would
+be a script that runs, utilizing the `generate_test_report` library to extract
+failure information, and then uploading the information to the web server.
+Information on how the premerge jobs will query the server and display results
+about flaky/already broken tests is in
+[the section below](#informing-the-user). We plan on having both the premerge
+and postcommit jobs upload failure information to the web server. This enables
+the web server to know about failures that are not the result of mid-air
+collisions before postcommit testing has been completed. Postcommit testing can
+take upwards of 30 minutes, during which the premerge advisor would not be able
+to advise that these failures are due to a broken `main`.
+
+We plan on implementing the web server in python with the `flask` library. All
+contributors to the premerge infrastructure are already familiar with Python,
+building web servers in Python with `flask` is relatively easy, and we do not
+need high performance or advanced features. For storage, we plan on using SQLite
+as it has support built in to Python, does not require any additional complexity
+in terms of infrastructure setup, and is reliable.
+
+Given we have two identical clusters that need to be able to function
+independently of each other, we also need to duplicate the web server for the
+premerge advisor. Queries and failure information uploads will go to the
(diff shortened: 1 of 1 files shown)


premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+
+The usefulness of the premerge system is significantly impacted by how much
+people trust it. People trust the system less when the issues reported to them
+are unrelated to the changes that they are currently working on. False positives
+occur regularly whether due to flaky tests, or due to main being broken. Efforts
+to fix these issues at the source and keep main always green and deflake tests
+are laudable, but not a scalable solution to the problem of false positive
+failures. It is also not reasonable to expect PR authors to spend time
+familiarizing themselves with all known flaky tests and dig through postcommit
+testing logs to see if the failures in their premerge run also occur in main.
```
Keenuts: nit/maybe out-of-topic: is that really unreasonable to look at failure logs and eyeball if we are responsible? Knowing which tests are flaky yes, but we should expect contributors to at least look at the logs if something fails

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+collisions before postcommit testing has been completed. Postcommit testing can
+take upwards of 30 minutes, during which the premerge advisor would not be able
+to advise that these failures are due to a broken `main`.
+
+We plan on implementing the web server in python with the `flask` library. All
+contributors to the premerge infrastructure are already familiar with Python,
+building web servers in Python with `flask` is relatively easy, and we do not
+need high performance or advanced features. For storage, we plan on using SQLite
+as it has support built in to Python, does not require any additional complexity
+in terms of infrastructure setup, and is reliable.
```
Keenuts, line 89: Could we consider going without the DB?
A DB, even SQLite means we have a schema, and if we change it, have a migration path.

What we need is history on the N last commits runs on main. Could this be stored in memory in the flask process? (assuming memory shared across the N workers if we use waitress)

It means if we restart the cluster, the first commit won't have history, but this shouldn't be that often and would simplify greatly incremental work on this advisor.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+
+Given we have two identical clusters that need to be able to function
+independently of each other, we also need to duplicate the web server for the
+premerge advisor. Queries and failure information uploads will go to the
+cluster-local premerge advisor web server. Periodically (eg once every thirty
+seconds), the web server will query the web server on the other cluster to see
+if there is any new data that has not been propgated back to the other side yet.
```
Keenuts:
```suggestion
if there is any new data that has not been propagated back to the other side yet.
```

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+contributors to the premerge infrastructure are already familiar with Python,
+building web servers in Python with `flask` is relatively easy, and we do not
+need high performance or advanced features. For storage, we plan on using SQLite
+as it has support built in to Python, does not require any additional complexity
+in terms of infrastructure setup, and is reliable.
+
+Given we have two identical clusters that need to be able to function
+independently of each other, we also need to duplicate the web server for the
+premerge advisor. Queries and failure information uploads will go to the
+cluster-local premerge advisor web server. Periodically (eg once every thirty
```
Keenuts: By not polling, and requesting each job to upload data to the service/DB, while also having clustering means we now need to handle data inconsistencies.

What about:
 - each job uploads an artifact to the github action storage (meaning retention policies etc is handled by Github).
 - one instance of this service is polling (as the metric container does), and fetches those artifacts as they are published.
 - the service keeps the test state in memory (no DB schema to migrate as we iterate on this service)
 - the service then posts a message to each PR as workflows are polled depending on the stored state.
 
 We could also have a single instance of this service as this is not a critical piece, a failure would just prevent sending helpful comments, but wouldn't back-up the queue.

boomanaiden154: We don't really need to handle data inconsistencies. The system is designed for eventual consistency on the order of minutes because having slightly inconsistent results doesn't really matter.

We can't use Github artifacts because we also need to support uploading data from the postcommit builders running on buildbot.

I think this is reasonably critical to the UX of the premerge system and not having two means we probably need to push maintenance to off-hours more often, but ack on looking more into keeping a single instance.

premerge/premerge-advisor.md
boomanaiden154: The majority of people already look at the logs. But you also need to be able to correlate the failures in your PR to postcommit failures, which is this missing piece. The heuristics I've seen most people use to determine if their PR is the culprit don't take into account whether or not there is a correlation and thus they are pretty error prone.

premerge/premerge-advisor.md
boomanaiden154: Whether we keep things in-memory or inside a DB, we still need a schema.  The schema that we would need is also not complicated. We just need the file, the workflow run number/commit SHA, the failure message, and maybe a timestamp/otherwise unique ID. Keeping the data for all of eternity also isn't that necessary. We can just dump the existing DB and start fresh, just like we would if we were storing info in memory and the server restarted.

Storing the information in memory makes it much harder to keep track of flaky tests over long periods of time, and we lose all information on flaky tests everytime the server restarts, which isn't ideal. Server restarts are normal and expected with k8s, so the design needs to take them into account.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
+# Premerge Advisor
+
+## Introduction
+
+While the premerge infrastructure is generally reliable, tests that are broken
```
cmtice: suggestion: "premerge infrastructure" => "premerge testing infrastructure"

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+at HEAD, or tests that are flaky can significantly degrade the user experience.
+Failures unrelated to the PR being worked on can lead to warning fatigue which
```
cmtice: suggestion: "Failures unrelated" => "Reporting failures unrelated"

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+can compound this problem when people land failing tests into main that they
```
cmtice: The entire phrase starting with "which can compound" needs to be reworked -- it is not clear to me what you are trying to say.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+believe are unrelated. The premerge advisor is designed to run after premerge
```
cmtice: I would suggest inserting a sentence: 'To address these issues we propose to implement a "premerge advisor".'  I'd put that just before the sentence that starts "The premerge advisor is..

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+testing and signal to the user whether the failures can be safely ignored because
+they are flaky or broken at head, or whether they should be investigated further.
+
+## Background/Motivation
+
+The usefulness of the premerge system is significantly impacted by how much
+people trust it. People trust the system less when the issues reported to them
+are unrelated to the changes that they are currently working on. False positives
+occur regularly whether due to flaky tests, or due to main being broken. Efforts
```
cmtice: suggested change: "main being broken" => "tests already being broken at head (perhaps by a previous commit)".

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+to fix these issues at the source and keep main always green and deflake tests
+are laudable, but not a scalable solution to the problem of false positive
+failures. It is also not reasonable to expect PR authors to spend time
```
cmtice: suggested changes:  "laudable, but not a scalable solution to"  => "ongoing, but are not sufficient to prevent" and  "failures." => "failure reports."

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+familiarizing themselves with all known flaky tests and dig through postcommit
+testing logs to see if the failures in their premerge run also occur in main.
+These alternatives are further explored in the section on
+[alternatives considered](#alternatives-considered). Having tooling to
+automatically run through the steps that one would otherwise need to perform
+manually would ensure the analysis on every failed premerge run is thorough and
+likely to be correct.
+
+## Design
```
cmtice, line 33: Your design doesn't say anything about how the system will notice when tests that were previously failing on main (at head) have been fixed.  If commit A breaks a test, commit B fixes the test, and commit C breaks the test again, you don't want the premerge advisor to tell commit C it can ignore the failure (because it still has the record that A broke it).

Also what if  two different commits break the same test, but in two different ways. Will the premerge advisor be smart enough to detect that the second commit caused a different breakage than the first commit?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+
+For the premerge jobs running through Github Actions, we plan on using the
+existing `generate_test_report` scripts that are currently used for generating
+summaries on job failures. When the job ends and there is a failure, there would
+be a script that runs, utilizing the `generate_test_report` library to extract
+failure information, and then uploading the information to the web server.
+Information on how the premerge jobs will query the server and display results
+about flaky/already broken tests is in
+[the section below](#informing-the-user). We plan on having both the premerge
+and postcommit jobs upload failure information to the web server. This enables
```
cmtice: It would be appropriate to include a link to the postsubmit testing docs.

cmtice: Also, I assume the data being uploaded from postcommit and PR jobs will be tagged and treated differently somehow (i.e. we want to store one but not the other).   More details about that would be good.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+ignored for that premerge run. If the failures are novel, then the web server
+can signal that the failures should be investigated more thoroughly. If there
+are failures, the premerge advisor can then write out a comment on the PR
+explaining its findings.
+
+### Processing and Storing Failures
+
+A key part of the premerge advisor is infrastructure to store and process build
+failure information so that it can be queried later. We plan on having jobs
+extract failure logs and upload them to a web server. This web server in
```
cmtice, line 55: Is this some existing webserver? Or is creating the webserver part of the premerge advisor implementation? More details about the implementation plans would be appreciated.

cmtice: Never mind; I just found the details below. :-)

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+addition to having an endpoint to accept uploads will have an endpoint that will
+accept test failure information (logs and filenames) and return whether or not
+they are broken at `main`, flaky, or novel test failures due to the PR.
+
+For the premerge jobs running through Github Actions, we plan on using the
+existing `generate_test_report` scripts that are currently used for generating
+summaries on job failures. When the job ends and there is a failure, there would
+be a script that runs, utilizing the `generate_test_report` library to extract
+failure information, and then uploading the information to the web server.
```
cmtice: nit "uploading" => "uploads"

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+uploaded to the web server. Afterwards, the premerge workflow then makes a
+request to the server asking if it can explain the failures as either existing
+in `main` or as flaky.
+
+After the response from the server has been recieved, the workflow will then
+construct a comment. It will contain the failure information, and if relevant,
+information/links showing the tests are flaky (and the flakiness rate) or are
+broken in `main`. If all of the test failures are due to flakes or failures in
+`main`, the comment will indicate to the user that they can safely merge their
+PR despite the test failures. We plan to construct the comment in a manner
```
cmtice: Is there any way to just mark the premerge test run as Passing, if all failure are due to flakes or failures at head?

cmtice: If there ever came a day when passing premerge tests became a requirement, it seems like being able to mark the tests as passing, when all failures are due to flakes or failures at head, would be a critical piece...

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+similar to the code formatting action. We will create one comment on the first
+workflow failure and then update that comment everytime we get new data. This
+prevents creating much noise. This does mean that the comment might get buried
+in long running reviews, but those are not the common case and people should
+learn to expect to look for the comment earlier in the thread in such cases.
```
cmtice: Shouldn't the comment about which test failures the user can ignore go into the Summary, at the very top?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Alternatives Considered
```
lnihlen, line 186: It looks like there are a number of off-the-shelf commercially available and open source flake detector systems. Here's a quote from Gemini:

A unit test flake detector identifies tests that pass and fail inconsistently without code changes by rerunning tests multiple times and analyzing the results for differing outcomes. These tools use historical data and CI/CD pipeline logs to flag unreliable tests, with features like [auto-rerunning](https://www.google.com/search?cs=0&sca_esv=6df9598451498e66&sxsrf=AE3TifOKuV5a57ZZr0SyuRorzFnSrvrBxw%3A1758728450252&q=auto-rerunning&sa=X&ved=2ahUKEwjQpO-D3vGPAxUEkYkEHaawDIgQxccNegQIBRAB&mstk=AUtExfDAng1Tye2_I7SSQFzqAmrJQ5RE81SH5Eu6qAi66rDrHfJv2JCRBs6vc18o1dPQh3DuTafii0QXC_jnLFys0V2cYnj6LhYB9UE4fv2-8OxXW7lQ4_kZlAvtuP788nfBAttfRHxK77Lg4i554gToVa8OtL_wrTtPC3jhWx32gYE5acM&csui=3), [historical analysis](https://www.google.com/search?cs=0&sca_esv=6df9598451498e66&sxsrf=AE3TifOKuV5a57ZZr0SyuRorzFnSrvrBxw%3A1758728450252&q=historical+analysis&sa=X&ved=2ahUKEwjQpO-D3vGPAxUEkYkEHaawDIgQxccNegQIBRAC&mstk=AUtExfDAng1Tye2_I7SSQFzqAmrJQ5RE81SH5Eu6qAi66rDrHfJv2JCRBs6vc18o1dPQh3DuTafii0QXC_jnLFys0V2cYnj6LhYB9UE4fv2-8OxXW7lQ4_kZlAvtuP788nfBAttfRHxK77Lg4i554gToVa8OtL_wrTtPC3jhWx32gYE5acM&csui=3), and [test quarantine](https://www.google.com/search?cs=0&sca_esv=6df9598451498e66&sxsrf=AE3TifOKuV5a57ZZr0SyuRorzFnSrvrBxw%3A1758728450252&q=test+quarantine&sa=X&ved=2ahUKEwjQpO-D3vGPAxUEkYkEHaawDIgQxccNegQIBRAD&mstk=AUtExfDAng1Tye2_I7SSQFzqAmrJQ5RE81SH5Eu6qAi66rDrHfJv2JCRBs6vc18o1dPQh3DuTafii0QXC_jnLFys0V2cYnj6LhYB9UE4fv2-8OxXW7lQ4_kZlAvtuP788nfBAttfRHxK77Lg4i554gToVa8OtL_wrTtPC3jhWx32gYE5acM&csui=3) for managing flagged tests. Popular examples include features within CI platforms like [Azure DevOps](https://www.google.com/search?cs=0&sca_esv=6df9598451498e66&sxsrf=AE3TifOKuV5a57ZZr0SyuRorzFnSrvrBxw%3A1758728450252&q=Azure+DevOps&sa=X&ved=2ahUKEwjQpO-D3vGPAxUEkYkEHaawDIgQxccNegQIBhAB&mstk=AUtExfDAng1Tye2_I7SSQFzqAmrJQ5RE81SH5Eu6qAi66rDrHfJv2JCRBs6vc18o1dPQh3DuTafii0QXC_jnLFys0V2cYnj6LhYB9UE4fv2-8OxXW7lQ4_kZlAvtuP788nfBAttfRHxK77Lg4i554gToVa8OtL_wrTtPC3jhWx32gYE5acM&csui=3), [Mergify](https://www.google.com/search?cs=0&sca_esv=6df9598451498e66&sxsrf=AE3TifOKuV5a57ZZr0SyuRorzFnSrvrBxw%3A1758728450252&q=Mergify&sa=X&ved=2ahUKEwjQpO-D3vGPAxUEkYkEHaawDIgQxccNegQIBhAC&mstk=AUtExfDAng1Tye2_I7SSQFzqAmrJQ5RE81SH5Eu6qAi66rDrHfJv2JCRBs6vc18o1dPQh3DuTafii0QXC_jnLFys0V2cYnj6LhYB9UE4fv2-8OxXW7lQ4_kZlAvtuP788nfBAttfRHxK77Lg4i554gToVa8OtL_wrTtPC3jhWx32gYE5acM&csui=3), and [TeamCity](https://www.google.com/search?cs=0&sca_esv=6df9598451498e66&sxsrf=AE3TifOKuV5a57ZZr0SyuRorzFnSrvrBxw%3A1758728450252&q=TeamCity&sa=X&ved=2ahUKEwjQpO-D3vGPAxUEkYkEHaawDIgQxccNegQIBhAD&mstk=AUtExfDAng1Tye2_I7SSQFzqAmrJQ5RE81SH5Eu6qAi66rDrHfJv2JCRBs6vc18o1dPQh3DuTafii0QXC_jnLFys0V2cYnj6LhYB9UE4fv2-8OxXW7lQ4_kZlAvtuP788nfBAttfRHxK77Lg4i554gToVa8OtL_wrTtPC3jhWx32gYE5acM&csui=3), as well as third-party tools like [Trunk.io](https://trunk.io/), [the flaky-tests-detection PyPI package](https://pypi.org/project/flaky-tests-detection/).

I think it's worth discussing the build/buy/re-use tradeoffs here with at least a few of these.

premerge/premerge-advisor.md
boomanaiden154: Added some text to describe what I'm trying to get at. Please take a look and let me know what you think.

premerge/premerge-advisor.md
boomanaiden154: Changed from laudable to ongoing.

I think it's still important to mention that this is a scalability problem. And fixing flaky tests fixes the failures (which is ideal) rather than just making them go unreported.

premerge/premerge-advisor.md
boomanaiden154: They're already required, it's just that everyone can opt-out by just clicking a check-box.

I've added some text on how we plan to achieve this.

premerge/premerge-advisor.md
boomanaiden154: We want to store both. But they will be tagged appropriately.

Added some text on how exactly we'll tag them. I don't think a design doc is the best place to get into exactly how the database schemas will be set up, but I can add that information if desired.

premerge/premerge-advisor.md
boomanaiden154: Added a new section on this (identifying if failures are novel).

The scenario that you described is possible and will trip up the premerge advisor, but there's no real way to alleviate it given the latency of postcommit testing. I've added an example.

I think the section is a bit terse for now and probably needs to be adjusted a bit more for clarity. If you have any specific pointers on what is hard to understand/where examples would be useful, that would be helpful.

premerge/premerge-advisor.md
boomanaiden154: Added a section to cover this. This actually ended up being a pretty useful exercise.

trunk.io looks almost exactly like what we want other than that it (along with all the other alternatives) only work with flaky tests rather than failures at head. We really need both to create a cohesive solution for LLVM.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+are failures, the premerge advisor can then write out a comment on the PR
+explaining its findings.
+
+### Processing and Storing Failures
+
+A key part of the premerge advisor is infrastructure to store and process build
+failure information so that it can be queried later. We plan on having jobs
+extract failure logs and upload them to a web server. This web server in
+addition to having an endpoint to accept uploads will have an endpoint that will
+accept test failure information (logs and filenames) and return whether or not
```
cmtice: What's the difference between accepting "uploads" and accepting "test failure information".  I think I know what is meant  (PR info vs postsubmit testing info) but that is not really clear here.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+For the premerge jobs running through Github Actions, we plan on using the
+existing `generate_test_report` scripts that are currently used for generating
+summaries on job failures. When the job ends and there is a failure, there would
+be a script that runs, utilizing the `generate_test_report` library to extract
+failure information, and then uploads the information to the web server.
+Information on how the premerge jobs will query the server and display results
+about flaky/already broken tests is in
+[the section below](#informing-the-user). We plan on having both the premerge
+and [postcommit jobs](post-submit-testing.md) upload failure information to the
+web server. This enables the web server to know about failures that are not the
```
cmtice: Maybe you should say something here about the data from the post commit jobs being stored in the DB; failures form the precommit jobs get compared against the DB to determine if the failures are new or already existed.

premerge/premerge-advisor.md
cmtice: You should explicitly state in the design doc why the inconsistencies won't be a problem. It would also be a good idea to outline the basic DB schema. Try to design it so that it's easily extensible if/when we decide we need to extend it.

cmtice: Also, if we're planning on using the DB to help identify flaky tests over time, which your comments imply, then that should be stated in the design doc as well.

cmtice: What was the upshot of "looking more into keeping a single instance"?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+for, and the other side can reply with the data that it has past that point.
+Explicitly synchronizing everytime without assumptions about the state of the
+other side has benefits over just writing through, like ensuring that a cluster
+that has been down for a significant amount of time is seamlessly able to
+recover. Synchronization does not need to be perfect as test failures that are
+flaky or broken at head will almost certainly show up in both clusters
+relatively quickly, and minor discrepancies for queries between the clusters are
+not a big deal.
+
+### Identifying if Failures are Novel
```
cmtice: I think this section needs a better name.  Maybe something like "Classifying Test Failures"?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+
+When the web server is sent a request to explain a list of failures, it needs to
```
cmtice, line 126: Is this the same thing as the PR just uploading its test results? Or is this actually a separate request?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+be able to determine whether a failure has occurred previously. To do this, the
+web server will keep a list of currently active failures in `main` and a list of
+flaky tests. The list of flaky tests will be computed from historical data on
+startup from a rolling window of data. The exact heuristic is left to
```
cmtice, line 130: Do you have a rough heuristic in mind that you could outline here?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+implementation as it will likely require some experimentation. The list of
+currently active failures will initially be taken from the last known postcommit
+run. If a new postcommit run shows additional failures that are not in the flake
+list, they will be added to the currently active failure list. Failures in the
+currently active list not present in the latest postcommit run will be removed
+from the currently active list. In addition to data from postcommit testing, if
+a PR lands with unexplained premerge failures, those failures are also added to
```
cmtice: What do you mean by "unexplained premerge failures" here?

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+Once a workflow has completed, no actions related to the premerge advisor will
+be performed if there are no failures. If there are failures, they will be
+uploaded to the web server. Afterwards, the premerge workflow then makes a
+request to the server asking if it can explain the failures as either existing
+in `main` or as flaky.
+
+After the response from the server has been recieved, the workflow will then
+construct a comment. It will contain the failure information, and if relevant,
+information/links showing the tests are flaky (and the flakiness rate) or are
+broken in `main`. If all of the test failures are due to flakes or failures in
```
cmtice: This sentence is nearly identical with the first sentence of the next paragraph. I think maybe you can omit it here.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+`main`, the comment will indicate to the user that they can safely merge their
+PR despite the test failures. We plan to construct the comment in a manner
```
cmtice: "comment" => "review comment" (maybe)?. Also, since (I think), we're also going to put this information into the Summary section, you should mention that.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+putting effort into deflaking tests and making process changes to ensure `main`
+is not broken. These fixes have the bonus of being useful for more than just
+premerge, also improving reliability for buildbots and any downstream testing.
+While we probably could achieve this at least temporarily with process changes
+and a concentrated deflaking effort, we do not believe this is feasible or
+scalable.
+
+In order to ensure that main is not broken by new patches, we need to ensure
+that every commit is tested directly on top of `main` before landing. This is
+not feasible given LLVM's current processes where pushing directly to main is a
```
cmtice: I think you should omit the sentence about direct pushes.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,252 @@
...
+critical component of several developer's workflows. We would also need to
+reduce the risk of "mid-air collisions", patches that pass premerge testing, but
+fail on `main` when landed due to the patch in its original state not being
+compatible with the new state of main. This would most likely involve merge
+queues which would introduce new CI load and are also not compatible with LLVM's
+existing practices for the same reason requiring premerge checks to be run
+before landing are not.
```
cmtice: I think you should omit the final phrase ("for the same reason requiring premerge checks to be run before landing are not").

premerge/premerge-advisor.md
boomanaiden154: I've added some text around why inconsistencies won't be a problem and also documented the current state (we have premerge write to both clusters rather than trying to have them sync between each other).

Also added some text on the database design.

premerge/premerge-advisor.md
boomanaiden154: As in the workflow summary page?

This section is specifically talking about adding comments mirroring that section. The info will be prominently placed within the comment, but the overall ordering of that thread relative to other comments in the PR will change as more comments get added.

premerge/premerge-advisor.md
boomanaiden154: Reworded. One is supposed to be just uploading failure information, and the other is supposed to be requesting explanations.

premerge/premerge-advisor.md
boomanaiden154: It's a separate request. `/explain` versus `/upload`.

premerge/premerge-advisor.md
boomanaiden154: Added some text.

premerge/premerge-advisor.md
boomanaiden154: Reworded this because the system as implemented does not work like this.

premerge/premerge-advisor.md
boomanaiden154: Added some text to cover this.

premerge/premerge-advisor.md
boomanaiden154: Done.

premerge/premerge-advisor.md
boomanaiden154: It's a normal comment, not a review comment.

It's also not going into the summary section currently. If/when it does, we can update the doc to match.

premerge/premerge-advisor.md
boomanaiden154: Is there a reason? I think it is an important reason for why we need postcommit testing on top of the others (like mid-air collisions).

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Alternatives Considered
+
+There are two main alternatives to this work. One would be to do nothing and let
+users figure out these failures on their own, potentially with documentation to
+better inform them of the process. The other alternative would be to work on
+keeping `main` green all the time and deflake tests rather than work on a
+premerge advisor. Both of these alternatives are considered below.
+
+### Deflaking Tests and Keeping Main Green
```
cmtice, line 194: Since this section appears to be describing something we considered but did not do (and are not planning on doing?), I would prefix the section title with "Alternative: ".  Just a suggestion.

boomanaiden154: It's a subtitle under the `Alternatives Considered` section, so it should be clear that it is nested within that section when actually rendered (although that does depend on the markdown renderer).

premerge/premerge-advisor.md
cmtice: Replied offline.

premerge/premerge-advisor.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Introduction
+
+While the premerge testing infrastructure is generally reliable, tests that are
+broken at HEAD, or tests that are flaky can significantly degrade the user
+experience. Reporting failures unrelated to the PR being worked on can lead to
+warning fatigue. Warning fatigue can self reinforce when people land failing
+tests into main that they believe are unrelated. This causes more false positive
+failures on other premerge testing invocations, leading to more warning
+fatigure. To address these issues we propose to implement a "premerge advisor".
```
dschuff:
```suggestion
fatigue. To address these issues we propose to implement a "premerge advisor".
```

premerge/premerge-advisor.md
boomanaiden154: Removed.

PR #604, opened 2025-10-07
[CI] Add Documentation on us-west Cluster Issue

boomanaiden154:
This patch adds some documentation on the issue impacting job scheduling on the us-west cluster.

### DIFF
--- a/premerge/issues.md
+++ b/premerge/issues.md
@@ -221,3 +221,61 @@ and the following for `llvm-premerge-cluster-us-west`:
 ```bash
 gcloud container clusters get-credentials llvm-premerge-cluster-us-west --location us-west1
 ```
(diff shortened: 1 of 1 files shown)


boomanaiden154:
> This looks great, as far as it goes. Do we have any idea how the configuration went from good to bad in the first place? Or how to prevent that in the future?

Not sure how it went from good to bad. There might be an audit log somewhere but I would suspect only Github admins have access. My guess is someone (me?) was playing around with the configuration and not realizing what they were doing.

Preventing it in the future is somewhat difficult as we cannot just lock down the configuration. We would need to use something like Terraform to manage the Github configuration or otherwise script an assert to ensure the configuration is as expected.

PR #605, opened 2025-10-07
[CI] Add Skeleton for Premerge Advisor Server

boomanaiden154:
This patch adds a skeleton for the premerge advisor server. Currently we just
add some endpoints with mostly dummy implementations. Having a skeleton enables
setting up the infrastructure to run the server and also adding infrastructure
in the monorepo to connect.


premerge/advisor/server.py
```diff
@@ -0,0 +1,21 @@
...
+
+import advisor_lib
+
+app = Flask(__name__)
+
+
+@app.route("/upload", methods=["POST"])
+def upload():
+    advisor_lib.upload_failures(flask.request.json)
+    return flask.Response(status=204)
```
cmtice, line 12: What is in the 'flask.request.json', and what does 'status=204' mean?

premerge/advisor/advisor_lib.py
```diff
@@ -0,0 +1,25 @@
...
+    message: str
+
+
+class FailureExplanation(TypedDict):
+    name: str
+    explained: bool
+    reason: str | None
+
+
+def upload_failures(test_failures: list[TestFailure]):
```
cmtice, line 15: Won't you need to distinguish between failures from PRs and failures from the postsubmit buildbot? Won't you want a function parameter for that?

boomanaiden154: Eventually, yes. This is just intended to be a skeleton so I can get the infrastructure up without needing a fully fleshed out implementation of the advisor.

Adding it now or later shouldn't make a big difference, so I can do whatever you prefer.

premerge/advisor/server.py
boomanaiden154: `flask.request.json` is just parsed JSON from the HTTP request that this route handles. `status=204` means respond with HTTP Status 204 which means ok/not returning any content. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

premerge/advisor/advisor_lib.py
cmtice: Ok as is for now then.

PR #607, opened 2025-10-07
[CI] Add workflow to run python unittests

boomanaiden154:
To ensure that they do not regress if someone forgets to run them locally.
This is similiar to the check-ci.yml worfklow in the monorepo.


.github/workflows/ci-python-unit.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: Check CI Python Unit Tests
+    if: github.repository_owner == 'llvm'
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v6
+      with:
+        python-version: '3.13'
+    - name: Run Tests
+      run: pytest premerge/ zorg/buildbot/builders/annotated/premerge/dispatch_job_test.py
```
cmtice, line 29: Based on the premerge testing :-), it looks like you may need to install pytest first?

boomanaiden154: Fixed. Meant to keep this marked as a draft until I got it working, but seems like you beat me to requesting a review. Sorry for the noise here.

.github/workflows/ci-python-unit.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  run_tests:
+    name: Check CI Python Unit Tests
+    if: github.repository_owner == 'llvm'
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v6
+      with:
+        python-version: '3.13'
```
DavidSpickett, line 21: 3.13 for any particular reason, is this being used server side?

boomanaiden154: Mostly just what I was generating `requirements.lock.txt` files with locally. This is what we plan on using server side  for the premerge advisor. Our buildbot container uses the python from Ubuntu 24.04, so 3.12, but should be close enough for the single script that we have.

PR #608, opened 2025-10-08
[test-suite] Install LNT using requirements file

DavidSpickett:
Instead of "setup.py develop". Fixes https://github.com/llvm/llvm-lnt/issues/76.

The new process is tested in CI so we can catch problems before they get to the buildbots.

https://llvm.org/docs/lnt/quickstart.html#installation

The bots only use the client so I've used that specific requirement file as suggested.

The requirement file has an entry ".". This means "install the package defined in the current directory". "current" means the current working directory, not the directory the requirement file is in.

This is handled by running the install command from within the LNT checkout. LNT will still be installed in the sandbox as expected.

### DIFF
--- a/zorg/buildbot/builders/ClangBuilder.py
+++ b/zorg/buildbot/builders/ClangBuilder.py
@@ -622,7 +622,7 @@ def _getClangCMakeBuildFactory(
                                description='install lnt dependencies',
                                workdir='test/sandbox',
                                env=env))
-        f.addStep(ShellCommand(name='install lit in sandbox',
+        f.addStep(ShellCommand(name='install LNT in sandbox',
                                command=[python, '-m', 'pip', 'install', '-r',
                                         'requirements.client.txt'],
                                haltOnFailure=True,


zorg/buildbot/builders/ClangBuilder.py
```diff
@@ -617,17 +616,18 @@ def _getClangCMakeBuildFactory(
                                description='recreating sandbox',
                                workdir='test',
                                env=env))
-        f.addStep(ShellCommand(name='install lnt dependencies',
+        f.addStep(ShellCommand(name='install setuptools in sandbox',
```
ldionne, line 619: Do you actually need to install setuptools?

DavidSpickett: Yes, see https://github.com/llvm/llvm-zorg/commit/fdc794ade01047608a01099ac702847e0948b383.

Not a problem for the current bots because they are still on the older version. Installing it is a NOP for them.

PR #611, opened 2025-10-10
[CI] Add deployments/services for premerge-advisor

boomanaiden154:
This allows for the premerge advisor server to run within the cluster and also
for the premerge advisor server container to be findable using the cluster's DNS
which allows us to query it/upload data from runner/buildbot pods.

### DIFF
--- a/premerge/premerge_resources/main.tf
+++ b/premerge/premerge_resources/main.tf
@@ -495,12 +495,12 @@ resource "kubernetes_namespace" "premerge_advisor" {
 }
 
 resource "kubernetes_manifest" "premerge_advisor_deployment" {
-  manifest   = yamldecode("advisor_deployment.yaml")
+  manifest   = yamldecode(file("advisor_deployment.yaml"))
   depends_on = [kubernetes_namespace.premerge_advisor]
 }
 
 resource "kubernetes_manifest" "premerge_advisor_service" {
-  manifest   = yamldecode("advisor_service.yaml")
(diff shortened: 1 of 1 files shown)


premerge/advisor_service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
cmtice, line 1: The name of this file seems odd; is  ".yaml.yml" really intended? Or was that some kind of typo?

premerge/premerge_resources/main.tf
```diff
@@ -488,6 +488,22 @@ resource "kubernetes_manifest" "libcxx_next_runners_disruption_budget" {
...
+  }
+}
+
+resource "kubernetes_manifest" "premerge_advisor_deployment" {
+  manifest   = yamldecode("advisor_deployment.yaml")
+  depends_on = [kubernetes_namespace.premerge_advisor]
+}
+
+resource "kubernetes_manifest" "premerge_advisor_service" {
+  manifest   = yamldecode("advisor_service.yaml")
```
cmtice: Should the name here exactly match the filename? (It doesn't at the moment, see comment above).

premerge/advisor_service.yaml
boomanaiden154: I created it with the wrong filename. Should be fixed now.

premerge/premerge_resources/main.tf
boomanaiden154: Renamed the filename, so this should match now.

PR #612, opened 2025-10-13
[CI] Refactor Premerge Advisor to use Application Factory

boomanaiden154:
Based on the flask documentation on the topic
(https://flask.palletsprojects.com/en/stable/patterns/appfactories/). This
refactoring enables us to add some integration tests to ensure that the
server plumbs all the data through correctly to the functions in
advisor_lib.py

### DIFF
--- a/premerge/advisor/advisor.py
+++ b/premerge/advisor/advisor.py
@@ -16,7 +16,7 @@ def explain():
     return advisor_lib.explain_failures(flask.request.json)
 
 
-def create_app(db_path: str):
+def create_app():
     app = Flask(__name__)
     app.register_blueprint(advisor_blueprint)
     return app


premerge/advisor/server.py
```diff
@@ -1,21 +1,6 @@
...
-
-
-@app.route("/explain")
-def explain():
-    return advisor_lib.explain_failures(flask.request.json)
+import advisor
 
 
 if __name__ == "__main__":
+    app = advisor.create_app()
```
cmtice, line 5: It looks like create_app requires an argument, but you haven't supplied one.

boomanaiden154: Fixed.

PR #614, opened 2025-10-13
[CI] Save Uploaded Failures to SQLite

boomanaiden154:
This patch makes the premerge advisor save failures to a SQLite database so
that we can query them later. We want to persist them to disk mainly to keep
a historical record of flaky tests.


premerge/advisor/advisor.py
```diff
@@ -1,13 +1,28 @@
...
 advisor_blueprint = flask.Blueprint("advisor", __name__)
 
 
+def _get_db():
+    if "db" not in flask.g:
+        flask.g.db = advisor_lib.setup_db(flask.current_app.config["DB_PATH"])
+    return flask.g.db
+
+
+def _close_db(exception):
```
cmtice, line 17: Do you need to handle the exception in some way?

premerge/advisor/advisor_lib.py
```diff
@@ -12,8 +14,39 @@ class FailureExplanation(TypedDict):
...
+    failures: list[TestFailure]
+
+
+def setup_db(db_path: str) -> sqlite3.Connection:
+    connection = sqlite3.connect(db_path)
+    tables = connection.execute("SELECT name from sqlite_master").fetchall()
+    if "failures" not in tables:
+        logging.info("Did not find failures table, creating.")
+        connection.execute(
+            "CREATE TABLE failures(source_type, base_commit_sha, source_id, test_file, failure_message)"
```
cmtice, line 30: Do we really want to create a failures table for failures from PRs? I thought we only wanted it for failures from postsubmit testing...

premerge/advisor/advisor.py
boomanaiden154: No. It only gets passed in if the request handler throws an exception. We want to unconditionally close the database, so don't need to do anything with it. It also already gets logged, so we shouldn't need to do anything else with it for now.

premerge/advisor/advisor_lib.py
boomanaiden154: We want to take information from both. Taking information from PRs allows us to respond more quickly if someone lands a PR that contains failures identified premerge.

PR #615, opened 2025-10-14
[CI] Add k8s infra for premerge advisor DB

boomanaiden154:
This patch adds PVCs and wires them up into the premerge advisor server pods
so that they can store their data persistently.

### DIFF
--- a/premerge/advisor_pvc.yaml
+++ b/premerge/advisor_pvc.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: advisor-db-pvc
-  namespacee: premerge-advisor
+  namespace: premerge-advisor
 spec:
   accessModes:
     - ReadWriteOnce


premerge/advisor_pvc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: advisor-db-pvc
+  namespacee: premerge-advisor
```
cmtice: typo: extra 'e' at end of 'namespace'.

Also, just wondering: Does it matter that on line 2 here "PersistentVolumeClaim" has a capital 'P', but on line 19 in the 'advisor_deployment.yaml' file, it's lowercase?

boomanaiden154: Fixed.

The capitalization is just the k8s syntax that someone decided at some point (probably with some reasoning behind it). The actual keys are all camel case, but `kind` is pascal case.

PR #616, opened 2025-10-16
[CI] Make Advisor Autoupdate Schema

boomanaiden154:
This patch makes the premerge advisor server autoupdate the DB schema if we
ever end up changing it. We handle this by just dropping the existing table
as preserving all data for all time is completely unnecessary in our case.
Unittests have also been added to ensure that the setup_db function works as
it is supposed to.

### DIFF
--- a/premerge/advisor/advisor_lib.py
+++ b/premerge/advisor/advisor_lib.py
@@ -1,4 +1,5 @@
 from typing import TypedDict
+import time
 import sqlite3
 import logging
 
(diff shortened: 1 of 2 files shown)


premerge/advisor/advisor_lib.py
```diff
@@ -21,15 +22,36 @@ class FailureUpload(TypedDict):
     failures: list[TestFailure]
 
 
+_CREATE_TABLE_CMD = "CREATE TABLE failures(source_type, base_commit_sha, source_id, test_file, failure_message)"
```
cmtice, line 25: Maybe add a comment that the "arguments" in the string below happen to be the same as the class variables above, but in fact they the names to be given to the table columns, and not the values of those variables.  If I got confused, other readers of the code could get confused too.

PR #617, opened 2025-10-16
[CI] Add Platform Information to Failure Results

boomanaiden154:
This patch adds platform information to the failure results. This is necessary
to disambiguate between failures on different architectures/operating systems.
The plan is to keep track of failing/flaky tests per platform, which this patch
enables.

### DIFF
--- a/zorg/jenkins/common.groovy
+++ b/zorg/jenkins/common.groovy
@@ -107,7 +107,7 @@ def benchmark_pipeline(label, body) {
 def testsuite_pipeline(label, body) {
     benchmark_pipeline(label) {
         dir('lnt') {
-            git url: 'https://github.com/llvm/llvm-lnt.git', branch: 'main', poll: false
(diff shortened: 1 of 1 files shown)


premerge/advisor/advisor_lib_test.py
```diff
@@ -70,6 +70,7 @@ def test_upload_failures(self):
                 {"name": "a.ll", "message": "failed in way 1"},
                 {"name": "b.ll", "message": "failed in way 2"},
             ],
```
cmtice, line 72: It looks like you're only testing for Linux failures? Wouldn't it be wise to test Windows failures as well?

boomanaiden154: Yes. We just pass the platform through directly to the database, so any string constant here would suffice and sufficiently test the functionality. I don't have any plans to do anything other than match strings for the platform. In cases where it matters I'll make sure both are tested, but here we do not gain any additional coverage by testing both.

PR #620, opened 2025-10-21
Add a new Linux worker for the pauth builders.

vvereschaka:


vvereschaka:
updated the devkit host names to more generic. Also reduced a number of available max jobs. We have VMs on the hosts and some CPUs have been reserved for them.

PR #623, opened 2025-10-23
collapse requests for ppc64le-flang bot

madanial0:
The [ppc64le-flang-rhel-clang](https://lab.llvm.org/buildbot/#/builders/157) bot has trouble keeping up with builds, queue was up to 13 days before clearing. Collapsing requests will help manage queue length.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2576,7 +2576,6 @@ all += [
 
     {'name' : 'ppc64le-flang-rhel-clang',
     'tags'  : ["flang", "ppc", "ppc64le"],
-    'collapseRequests' : False,
     'workernames' : ['ppc64le-flang-rhel-test'],
     'builddir': 'ppc64le-flang-rhel-clang-build',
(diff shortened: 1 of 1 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -2576,7 +2576,6 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
 
     {'name' : 'ppc64le-flang-rhel-clang',
     'tags'  : ["flang", "ppc", "ppc64le"],
-    'collapseRequests' : False,
```
DanielCChen, line 2579: Nit: Instead of deleting the line, would it be more beneficial to change the setting to `true`? So in the future, if we ever want to flip it, we can just change it back.

kkwli (approved):
LG

buildbot/osuosl/master/config/builders.py
gkistanova: No. Mark did it right. The correct way is to let the default machinery do its job.
Setting it to True will force collapsing build requests that should not be collapsed, like manual builds and such.

PR #624, opened 2025-10-23
[CI] Add Library for Mapping Commit SHAs to Indexes

boomanaiden154:
We need to know the index of commits to make some heuristics easier to
write, for example, are there any reported failures in the last five
commits. This patch adds a library that allows getting the index of a
commit assuming linear history. The plan is to use this to ensure the
failures database has commit indices rather than commit SHAs.

### DIFF
--- a/premerge/advisor/git_utils.py
+++ b/premerge/advisor/git_utils.py
@@ -46,9 +46,14 @@ def _get_and_add_commit_index(
         check=True,
     )
     log_lines = log_output.stdout.decode("utf-8").split("\n")[:-1]
-    for line_index, log_line in enumerate(log_lines):
+    if len(log_lines) == 0:
+        raise ValueError(
+            "Did not find any commits. The commit likely happened before the commit with index 1."
+        )
+    commit_index = latest_index + len(log_lines) + 1
+    for log_line in log_lines:
         line_commit_sha = log_line.split(" ")[0]
-        commit_index = latest_index + len(log_lines) - line_index
+        commit_index -= 1
         commits_to_add.append((line_commit_sha, commit_index))
     db_connection.executemany("INSERT INTO commits VALUES(?, ?)", commits_to_add)
     if not latest_commit_info:
--- a/premerge/advisor/git_utils_test.py
+++ b/premerge/advisor/git_utils_test.py
@@ -125,3 +125,20 @@ class GitUtilsTest(unittest.TestCase):
             ),
             4,
         )
+
+    def test_get_index_error_invalid_sha(self):
+        commit_shas = self.setup_repository(3)
+        with self.assertRaises(ValueError):
+            git_utils.get_commit_index(
+                commit_shas[0],
+                self.repository_path.name,
+                self.db_connection,
+                commit_shas[1],
+            )
+
+    def test_get_index_error_before_first_commit(self):
+        commit_shas = self.setup_repository(3)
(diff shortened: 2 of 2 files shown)


premerge/advisor/advisor_lib.py
```diff
@@ -28,13 +28,15 @@ class TestExplanationRequest[TypedDict]:
...
+_TABLE_SCHEMAS = {
+    "failures": "CREATE TABLE failures(source_type, base_commit_sha, source_id, test_file, failure_message, platform)",
+    "commits": "CREATE TABLE commits(commit_sha, commit_index)"
+}
 
 
 def _create_failures_table(connection: sqlite3.Connection):
     logging.info("Did not find failures table, creating.")
-    connection.execute(_CREATE_TABLE_CMD)
+    connection.execute(_CREATE_FAILURES_TABLE_CMD)
```
cmtice: I don't see _CREATE_FAILURES_TABLE_CMD defined anywhere?

boomanaiden154: Fixed. Accidentally mixed changes between patches.

premerge/advisor/git_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    _clone_repository_if_not_present(repository_path)
+    # Check to see if we already have the commit in the DB.
+    commit_matches = db_connection.execute(
+        "SELECT * FROM commits WHERE commit_sha=?", (commit_sha,)
+    ).fetchall()
+    if len(commit_matches) > 1:
+        raise ValueError("Expected only one entry per commit SHA")
+    elif len(commit_matches) == 1:
+        return commit_matches[0][1]
+    # We have not seen this commit before. Count the index and then add it to
```
cmtice, line 79: Is there any possibility that someone might pass in a commit sha that was committed previous to FIRST_COMMIT_SHA? If so, what happens?

premerge/advisor/git_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    else:
+        latest_sha = first_commit_sha
+        latest_index = 1
+    log_output = subprocess.run(
+        ["git", "log", "--oneline", "--no-abbrev", f"{latest_sha}..{commit_sha}"],
+        cwd=repository_path,
+        stdout=subprocess.PIPE,
+        check=True,
+    )
+    log_lines = log_output.stdout.decode("utf-8").split("\n")[:-1]
```
cmtice, line 48: Just wondering: Why are you omitting the last line ? ([:-1])

premerge/advisor/git_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    for line_index, log_line in enumerate(log_lines):
+        line_commit_sha = log_line.split(" ")[0]
+        commit_index = latest_index + len(log_lines) - line_index
```
cmtice: It would probably be simpler to initialize commit_index to latest_index just above the loop, then increment it inside the loop...

premerge/advisor/git_utils.py
boomanaiden154: The last line also has a `\n` at the end of the last string, so the last string in the split is just "" which we don't care about.

premerge/advisor/git_utils.py
boomanaiden154: Good point. I think currently we end up just erroring out, but maybe in a bad spot. I've added an explicit check and test coverage.

PR #627, opened 2025-10-24
[CI] Setup Networking for Premerge Advisor

boomanaiden154:
This patch sets up cross cluster networking for the premerge advisor. Primarily,
we add tags to the service pool instances to allow targeting them in a firewall
rule, disable firewall rule reconciliation on the cluters so that we can setup
custom ingress rules denying access to the general internet. We also add the
actual firewall rule allowing ingress.

### DIFF
--- a/premerge/advisor/advisor_lib.py
+++ b/premerge/advisor/advisor_lib.py
@@ -29,36 +29,40 @@ class TestExplanationRequest[TypedDict]:
     platform: str
 
 
-_CREATE_TABLE_CMD = "CREATE TABLE failures(source_type, base_commit_sha, source_id, test_file, failure_message, platform)"
+_TABLE_SCHEMAS = {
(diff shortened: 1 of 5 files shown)


premerge/gke_cluster/main.tf
```diff
@@ -44,6 +45,8 @@ resource "google_container_node_pool" "llvm_premerge_linux_service" {
     workload_metadata_config {
       mode = "GKE_METADATA"
     }
+
+    tags = ["premerge-service"]
```
cmtice, line 49: Don't we also need this for Windows?

boomanaiden154: No. This is specifically for the service pool where the premerge advisor runs. We don't need permissions in the node pools that run the workers because they'll only be accessing the cluster local premerge advisor.

cmtice (approved):

PR #630, opened 2025-10-24
Add extended factory function for TestSuiteBuilder factories.

vvereschaka:
Added an initial version of TestSuiteBuilder.getLlvmTestSuiteSteps() that is used to prepare the build steps to retrieve, build and run the LLVM TestSuite tests. This factory supports the CMake buildbot step, synchronization and run the tests on the remote boards and could be used together with the composite factories, such as UnifiedTreeBuilder.getCmakeExBuildFactory().

This factory is planned as a replacement for the current factory later.


vvereschaka:
This factory is going to be used with the llvm-clang-ubuntu-x-aarch64-pauth builder initially (#629).
We need these changes only on the staging master for now.

PR #632, opened 2025-10-24
[CI] Wire Up Premerge Advisor Hosts

boomanaiden154:
This patch passes the hostname of the other premerge advisor into the container
so that we know where to request information that might exist on the other cluster.
The IP addresses are stable with how GKE allocates them for external load
balancers. Eventually we probably want to switch over to domains, but that's not
necessary for now.


boomanaiden154:
Actually, planning on doing information exchange differently, so closing.

PR #639, opened 2025-10-30
[CI] Explain More Failures at HEAD

boomanaiden154:
This patch updates _try_explain_failing_at_head to also look at the
couple commits behind the base SHA as well. This accounts for the case
where we do not yet have postcommit results yet for the exact base
commit. This should not reduce accuracy given we're matching the failure
message exactly.

### DIFF
--- a/premerge/advisor/advisor_lib.py
+++ b/premerge/advisor/advisor_lib.py
@@ -115,7 +115,7 @@ def _try_explain_failing_at_head(
         )
     else:
         query += "AND base_commit_sha=?"
-        query_params += base_commit_sha
+        query_params += (base_commit_sha,)
     test_name_matches = db_connection.execute(
         query,
         query_params,
--- a/premerge/advisor/advisor.py
+++ b/premerge/advisor/advisor.py
@@ -30,7 +30,9 @@ def upload():
 
 @advisor_blueprint.route("/explain")
 def explain():
-    return advisor_lib.explain_failures(flask.request.json, _get_db())
+    return advisor_lib.explain_failures(
+        flask.request.json, flask.current_app.config["REPO_PATH"], _get_db()
+    )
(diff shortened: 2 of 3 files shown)


premerge/advisor/advisor_lib.py
```diff
@@ -95,12 +97,28 @@ def _try_explain_failing_at_head(
...
+    query = (
         "SELECT failure_message FROM failures "
-        "WHERE source_type='postcommit' AND base_commit_sha=? AND platform=?",
-        (base_commit_sha, platform),
+        "WHERE source_type='postcommit' AND platform=?"
+    )
+    query_params = (platform,)
+    if base_commit_index:
+        min_commit_index = (
+            base_commit_index - EXPLAINED_HEAD_MAX_COMMIT_INDEX_DIFFERENCE
```
cmtice, line 110: What if this subtraction generates a negative number?

boomanaiden154: It doesn't end up impacting any functionality. It basically just implies there is no restriction on the lower bound of the commit.

premerge/advisor/advisor_lib.py
```diff
@@ -95,12 +97,28 @@ def _try_explain_failing_at_head(
...
+        )
+        query += (
+            f" AND commit_index > {min_commit_index} "
+            f"AND commit_index <= {base_commit_index}"
+        )
+    else:
+        query += "AND base_commit_sha=?"
+        query_params += (base_commit_sha,)
```
cmtice, line 118: It looks like query_params is a tuple? In which case I would not expect "+=" to be a valid operator?

cmtice: Never mind; I've double-checked and this works.

PR #641, opened 2025-11-01
[CI] Explain Flaky Failures

boomanaiden154:
This patch adds functionality to the premerge advisor to explain away
flaky failures. For this, we just look to see that the failure observed
is failing for a sufficient amount of time, in our case 200 commits.


premerge/advisor/advisor_lib.py
```diff
@@ -131,20 +132,57 @@ def _try_explain_failing_at_head(
     return None
 
 
+def _try_explain_flaky_failure(
```
dwblaikie, line 135: Maybe a comment describing the particular definition of "flaky" this function is using would be handy?

It looks to me like it's testing if this failure has been seen before over at least a 200 commit range between oldest/newest sighting? What's to stop that being "it's been failing at HEAD for at least 200 revisions"?

boomanaiden154: Added a function level docstring.

I've never seen a non-flaky test failure that has been failing for more than ~20 commits. I think a OOM more in range should be safe. I thought about also adding a heuristic around the longest running subsequence to catch cases like this, but I think it adds complexity for little benefit:
1. As mentioned, 200 is about an OOM bigger than the ranges we've seen in practice.
2. We're exactly matching the failure message, so the test would need to fail in the exact same way to trigger this.

dwblaikie: I'd probably still go for a test that verifies the failures are discontiguous from ToT rather than a heuristic that guesses that something seems unlikely to be a long-standing failure at HEAD - seems cheap/simple enough to do. Because bad advice can be pretty confusing.

But I guess it's a start.

boomanaiden154: Yeah, I'll add in that functionality soon. It would be good to have it for the failing at head case too.

PR #648, opened 2025-11-06
Introduce ScriptedBuilder

Meinersbur:
Allow the main llvm-project repository to contain the buildbot builder instructions, instead of storing them in llvm-zorg.  The corresponding llvm-project PR is https://github.com/llvm/llvm-project/pull/166809.

Using polly-x86_64-linux-test-suite as a proof-of-concept because that builder is currently offline, I am its maintainer, and is easier to build than an configuration supporting offload. Once the design has been decided, more builders can follow.

Advantages are:
 * It is easier to make changes in the llvm-project repository. There are more reviewers than for the llvm-zorg repository.
 * Buildbot changes can be made in the same PR with changes that require updating the buildbot, e.g. changing the name of a CMake option.
 * Configuration changes take effect immeditately when landing; no buildbot master restart needed.
 * Some builders store a CMake [cache](https://github.com/llvm/llvm-project/tree/main/lldb/cmake/caches) [file](https://github.com/llvm/llvm-project/tree/main/lldb/cmake/caches) [in the](https://github.com/llvm/llvm-project/tree/main/libcxx/cmake/caches) [llvm-project repository](https://github.com/llvm/llvm-project/tree/main/clang/cmake/caches) for the reasons above. However, the number of changes that can be made with a CMake cache file alone are limited

Compared to AnnotatedBuilder, advantages are:
 * Reproducing a buildbot configuration locally made easy: just execute the script in-place. No llvm-zorg, local buildbot worker, or buildbot master needed. 
 * Same for testing a change of a builder before landing it in llvm-zorg. Doing so with an AnnotatedBuilder requires [two llvm-zorg checkouts](https://jplehr.de/2025/06/02/testing-and-developing-llvm-buildbots/): One for making the change of the builder script itself, which then is pushed to a private llvm-zorg branch on GitHub, and a second that is modified to fetch that branch instead of https://github.com/llvm/llvm-zorg/tree/main. 
 * The AnnotatedBuilder scripts are located in the llvm-zorg repository and the buildbot-workers always checkout is always the top-of-trunk. This means that a buildbot configuration is split over three checkouts:
     * The checkout of llvm-project to be tested
     * The checkout of llvm-zorg by the buildbot-worker fetches; always the top-of-trunk, i.e may not match the revision of llvm-project that is executed (such as the CMake cache files located there), especially when using the "Force build" feature.
     * The checkout of llvm-zorg that the buildbot-master is running, which is updated only when the master is manually restarted.
 * The "Force Build" feature also allows for test-building any llvm-project PR. This is correctly handled by zorg's `addGetSourcecodeSteps`, but does not work with AnnotatedBuilders that checkout the llvm-project source on their own.
     
The goal is to move as much as possible into the llvm-project repository such that there cannot be a mismatch between checkouts of different repositories. Ideally, the buildbot-master only needs to be updated+restarted for adding/removing workers, not for build configuration changes.

This has been discussed in the [Bi-Weekly LLVM Offload Meeting (Agenda item 13)](https://docs.google.com/document/d/1PAeEshxHCv22JDBCPA9GXGggLp0t7rsnD_jL04MBbzw/edit?tab=t.0#heading=h.d722ra7aauiu). There were no concerns.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -3,7 +3,6 @@ from importlib import reload
 from buildbot.plugins import util, steps
 
 from zorg.buildbot.builders import ClangBuilder
-from zorg.buildbot.builders import ScriptedBuilder
 from zorg.buildbot.builders import FlangBuilder
 from zorg.buildbot.builders import PollyBuilder
 from zorg.buildbot.builders import LLDBBuilder
@@ -20,6 +19,7 @@ from zorg.buildbot.builders import XToolchainBuilder
 from zorg.buildbot.builders import TestSuiteBuilder
 from zorg.buildbot.builders import BOLTBuilder
 from zorg.buildbot.builders import DebugifyBuilder
+from zorg.buildbot.builders import ScriptedBuilder
 
 from zorg.buildbot.builders import HtmlDocsBuilder
 from zorg.buildbot.builders import DoxygenDocsBuilder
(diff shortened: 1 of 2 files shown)


Meinersbur:
> Why this over an annotated builder with the script in zorg?

The other part is in https://github.com/llvm/llvm-project/pull/166809 which includes the motivation. Both are currently explicitly in draft mode to collect feedback

Meinersbur:
Reconnected polly-x86_64-linux-test-suite to staging: https://lab.llvm.org/staging/#/builders/49

Will see the change when the master updates

PR #649, opened 2025-11-06
[CI] Make Container Build Jobs use LLVM Actions

boomanaiden154:
Some common actions in https://github.com/llvm/llvm-project/pull/166663 were recently introduced to make building and pushing containers much simpler. Migrate the jobs in zorg to use them to avoid the duplication.


boomanaiden154:
Depends on https://github.com/llvm/llvm-project/pull/166864.

PR #653, opened 2025-11-11
Add helper function for the buildbot tests into zorg.buildbot.tests.

vvereschaka:
Also added the initial tests for the TestSuiteBuilder.getLlvmTestSuiteSteps factory.

### DIFF
--- a/test/buildbot/builders/testsuite.py
+++ b/test/buildbot/builders/testsuite.py
@@ -22,15 +22,14 @@ f = TestSuiteBuilder.getLlvmTestSuiteSteps(
         hint = None,
     )
 
+print(f"default factory: {f}\n")
+
 assert factory_has_num_steps(f, 5)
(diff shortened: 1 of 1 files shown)


test/buildbot/builders/testsuite.py
```diff
@@ -0,0 +1,51 @@
...
+
+assert factory_has_num_steps(f, 5)
+assert factory_has_step(f, "clean-src-dir")
+assert factory_has_step(f, "cmake-configure")
+assert factory_has_step(f, "build-default")
+assert not factory_has_step(f, "rsync-default")
+assert factory_has_step(f, "test-check")
+
+
+print(f"default factory: {f}\n")
```
slydiman: I would place this print before any asserts.

vvereschaka: yes, agree, thank you. Updated accordingly.

PR #657, opened 2025-11-14
[AIX] increase the number of jobs used for building AIX builds

madanial0:
Adding more jobs to aix workers as the machine utilization allows it and aix bots take a long time to complete builds

### DIFF
--- a/buildbot/osuosl/master/config/workers.py
+++ b/buildbot/osuosl/master/config/workers.py
@@ -116,8 +116,8 @@ def get_all():
         create_worker("sanitizer-buildbot12", properties={'jobs': 48}, max_builds=2),
 
         # POWER 8 PowerPC AIX 7.2
-        create_worker("aix-ppc64", properties={'jobs': 10}, max_builds=1),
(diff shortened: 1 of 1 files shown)


buildbot/osuosl/master/config/workers.py
```diff
@@ -116,8 +116,8 @@ def get_all():
         create_worker("sanitizer-buildbot12", properties={'jobs': 48}, max_builds=2),
 
         # POWER 8 PowerPC AIX 7.2
-        create_worker("aix-ppc64", properties={'jobs': 10}, max_builds=1),
-        create_worker("ppc64-flang-aix-test", properties={'jobs': 10}, max_builds=1),
+        create_worker("aix-ppc64", properties={'jobs': 20}, max_builds=1),
```
daltenty, line 119: nit: if this is the `-j` value I wonder if it can just be removed and let the build run full width.

madanial0: Both AIX workers are on the same machine not sure how that will behave

PR #660, opened 2025-11-15
Collapse buildrequests regardless of clean(_obj) property

Meinersbur:
With this PR, the `clean` and `clean_obj` properties are ignored when considering whether to collapse two buildrequests. The collapsed request will have that property if either of the original requests has. `clean`, respectively `clean_obj`, not being set never guaranteed pre-populated build directory anyway.

Background:
 * The `clean` property is set in the Force Build dialog: "Clean source code and build directory". It is never set by LLVMPoller.
 * The `clean_obj` property is set in the Force Build dialog: "Clean build directory", or when LLVMPoller adds a buildrequest for a new llvm-project commit that [changes a CMake file](https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/changes/llvmgitpoller.py#L170) or contains the string ["Require[...]clean build"](https://github.com/llvm/llvm-zorg/blob/main/zorg/buildbot/changes/llvmgitpoller.py#L40) in the commit message.
 * Neither is automatically set on builders even with [`clean` option](https://github.com/llvm/llvm-zorg/blob/main/buildbot/osuosl/master/config/builders.py#L65). Instead, those builders enable the cleaning step independent of the `clean`/`clean_obj` properties inherited from the buildrequest.

Two buildrequests were only collapsed when their properties are equal, including the `clean_obj` property. Since the value of the property changes dependending on whether a CMakeLists.txt file is touched, a typical sequence of buildrequests after collapsing becomes

1. buildrequest 
2. buildrequest with clean_obj=true
3. buildrequest 
4. buildrequest with clean_obj=true
5. buildrequest 
6. buildrequest with clean_obj=true
7. buildrequest 
8. ...

Hence, slow builders with `collapseRequests` enabled in practice it still does two builds per CMakeLists.txt change. This may lead to high response latency during times of high commit volume or the slowest builders not being able to keep up at all.

### DIFF
--- a/zorg/buildbot/process/buildrequest.py
+++ b/zorg/buildbot/process/buildrequest.py
@@ -31,7 +31,7 @@ def setBuildsetProperty(db, bsid, name, value, source="Collapse"):
     yield db.pool.do(thd)
 
     # Also update the lookup cache, if this buidset properties' has been cached.
-    if bsid in db.buildsets. getBuildsetProperties.cache.keys():
+    if bsid in db.buildsets.getBuildsetProperties.cache.keys():
         # Lookup of old values will be from the cache
         properties = yield db.buildsets.getBuildsetProperties(bsid)
 
--- a/zorg/buildbot/process/buildrequest.py
+++ b/zorg/buildbot/process/buildrequest.py
@@ -1,4 +1,45 @@
 from twisted.internet import defer
+import json
+import sqlalchemy as sa
+
+
+@defer.inlineCallbacks
+def setBuildsetProperty(db, bsid, name, value, source="Collapse"):
+    """Set a buildset property
+
+    buildbot.db.buildset.BuildsetConnectorComponent only has
+    getBuildsetProperties, but no setter. This setter is modelled after
+    setBuildProperty.
+    """
+    def thd(conn):
+        bs_props_tbl = db.model.buildset_properties
+        db.buildsets.checkLength(bs_props_tbl.c.property_name, name)
+
+        whereclause=sa.and_(bs_props_tbl.c.buildsetid == bsid, bs_props_tbl.c.property_name == name)
+        q = sa.select([bs_props_tbl.c.property_name, bs_props_tbl.c.property_value], whereclause=whereclause)
+        prop = conn.execute(q).fetchone()
+        value_js = json.dumps([value,source])
+        if prop is None:
+            conn.execute(bs_props_tbl.insert(), {
+                "buildsetid": bsid,
+                "property_name": name,
+                "property_value": value_js
+            })
+        elif prop.property_value != value_js:
+            conn.execute(bs_props_tbl.update(whereclause=whereclause), {"property_value": value_js})
+
+    yield db.pool.do(thd)
+
+    # Also update the lookup cache, if this buidset properties' has been cached.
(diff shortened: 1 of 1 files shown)


DavidSpickett:
My original PR for reference - https://github.com/llvm/llvm-zorg/pull/266/

Meinersbur:
What this does:

1. When considering collapsing merges, ignore `clean` and `clean_obj` properties.
2. The request are merged, set `clean`/`clean_obj` on the merged buildrequest, if at least one of the buildrequest has that property set.

This is different from the mechnism that #266 introduced in that it is able to set the `clean`/`clean_obj` property itself, whereas #266 only handles `clean_obj`  and requires that another component (in this case: `builder.config.factory.clean`) already applies that same effect as `clean_obj` would. Hence, it only works if `builder.config.factory.clean == True` (i.e. incremental builds disabled). This PR also works for incremental builders.

This PR also ignores the buildrequest `clean` property (that #266 still requires to match) which conceptually makes no difference here, but I don't have a use case. I still implemented it following the rule of least astonishment.

~~I think that #266 does not work as intended. The `clean` property is basically never set, nor do we want it to be set as it means that `git clone` has to re-populate the llvm.src directory. This means the condition `if getattr(builder.config.factory, "clean", False):` never applies.~~

Since it checks `builder.config.factory` (not the buildrequest properties), it actually might work for Factories that support it. The interpretation of `clean` is different than the buildrequest `clean` here.  AnnotatedBuilder, which I am working with, uses the term `clobber` instead, which might be why I didn't notice it actally collapsing.


> which seems to do the same thing as clean_obj?

 * Dialog description of `clean`: "Clean source code and build directory"
 * Dialog description of `clean_obj`: "Clean build directory"

Please also read the background section in the summary.

zorg/buildbot/process/buildrequest.py
```diff
@@ -1,4 +1,45 @@
...
+                "property_name": name,
+                "property_value": value_js
+            })
+        elif prop.property_value != value_js:
+            conn.execute(bs_props_tbl.update(whereclause=whereclause), {"property_value": value_js})
+
+    yield db.pool.do(thd)
+
+    # Also update the lookup cache, if this buidset properties' has been cached.
+    if bsid in db.buildsets. getBuildsetProperties.cache.keys():
```
DavidSpickett: Is the space after `buildsets.` intentional?

(I didn't realise Python let you do this, but it does)

DavidSpickett (commented):
I understand this now. A lot of my confusion was that my previous change definitely did do something for some of the builders, but I do now see how this PR expands on that.

Looks fine to me, but since this interacts with the database, Galina should be the approver.

And for them, the inspiration for the new setter method is here - https://github.com/buildbot/buildbot/blob/0aee5e842fcbb8b98e90fce01ea871e675445ce8/master/buildbot/db/builds.py#L320

zorg/buildbot/process/buildrequest.py
Meinersbur: Not intentional, fixed.

I get such spaces often from code completion thinking the formatter will remove those later. However, llvm-zorg is not pre-formatted, so running `darker` (especially on buildbot/osusol/master/config/builders.py) gets you a massive amount of spurious changes.

gkistanova (approved):
LGTM.

PR #662, opened 2025-11-18
[AIX] limit lit tests on clang builder to avoid limiting other builders

madanial0:
The machine that hosts clang aix builder also hosts the flang aix builder. Limit the number of threads the clang build uses in order not to drought the flang build as both run at the same time.


daltenty (approved):
LGTM

PR #668, opened 2025-12-04
Add a new SYCL GPU builder

sarnex:
Add a new Intel-hosted machine with a GPU that builds the `libsycl` project.

Once we have tests, we will run the `check` target for `libsycl` as well which will include tests that use the GPU.


gkistanova (commented):
LGTM.

PR #670, opened 2025-12-09
[libc++] Retrieve the Docker image used by libc++ pre-merge runners from the monorepo

ldionne:
This allows storing the various images associated to runner sets for libc++ within the monorepo, which greatly simplifies maintenance and synchronization on the libc++ side.


premerge/premerge_resources/variables.tf
```diff
@@ -59,21 +59,6 @@ variable "runner_group_name" {
   type = string
 }
 
-variable "libcxx_runner_image" {
```
ldionne, line 62: I would much rather leave them in `variables.tf`, but apparently the default value of a variable can't be something dynamic, so I had to "downgrade" those to `locals`.

boomanaiden154 (approved):

PR #673, opened 2025-12-10
Use LLVM_ENABLE_RUNTIMES for building compiler-rt in Green Dragon

ndrewh:
Building compiler-rt with LLVM_ENABLE_PROJECTS is deprecated and will be removed in a future release. This builds compiler-rt with LLVM_ENABLE_RUNTIMES.

Note that if any of these builds are used for release branches that do not yet have https://github.com/llvm/llvm-project/pull/166837, we may want to either cherry-pick that to the release branches or hold off on this PR.

### DIFF
--- a/zorg/buildbot/builders/sanitizers/buildbot_qemu.sh
+++ b/zorg/buildbot/builders/sanitizers/buildbot_qemu.sh
@@ -25,8 +25,7 @@ readonly STAGE2_DIR=llvm_build2_host
 
   cmake \
     ${CMAKE_COMMON_OPTIONS} \
-    -DLLVM_ENABLE_PROJECTS="clang;lld" \
-    -DLLVM_ENABLE_RUNTIMES="compiler-rt" \
+    -DLLVM_ENABLE_PROJECTS="clang;compiler-rt;lld" \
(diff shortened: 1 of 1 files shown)


zorg/buildbot/builders/sanitizers/buildbot_qemu.sh
azharudd, line 1: Please skip the changes to this file from this PR. Only the files in `jenkins` directories are applicable to Green Dragon.

DavidSpickett: Also add to the title "in Green Dragon". (as there are some Buildbot bots that are not using runtimes yet)

DavidSpickett: ~~I think this script was added by @asb so I assume they will change it when they're ready.~~

Actually I'm thinking of the *other* qemu script, sorry for the unnecessary ping.

ndrewh:
The fix for builtins tests got rolled back, so I'm waiting on the reland (https://github.com/llvm/llvm-project/pull/171941) before I merge this.

ndrewh:
This broke some jobs, fixing

PR #677, opened 2025-12-12
Add artifact manager script for downloading toolchains

justice-adams-apple:
* Add `artifact_manager` for downloading toolchain artifacts used by stage 2 builds.
* Add `fetchArtifact` shared library function for using the script in jenkins jobs

This will replace the fetch functionality in `monorepo_build.py` and be used by all clang jobs in the future, both the mainline (standard post merge jobs which exist now) and the bisection jobs (jobs which will be setup to mimic the mainline jobs in a separate folder) will use this script.

The script has the functionality to search for a stage1 build of both the mainline and bisection version of the job if needed and download the appropriate package. This saves on build time as we can utilize already built stage 1 packages if they exist. If no stage 1 artifact exists, the data is written to a props file and the CI system will fire off a stage 1 build accordingly.

This handles all of the logic around determining which toolchain package to download making it simple for CI libraries in the future to invoke the script as such:

By extracting the logic to the python scripts we can make the functionality portable across CI systems.

end users will call `fetchArtifact` like so

```
fetchArtifact([
    stage1Job: "clang-stage1-RA"
])
```

### DIFF
--- a/vars/fetchArtifact.groovy
+++ b/vars/fetchArtifact.groovy
@@ -5,7 +5,7 @@ def call(Map config = [:]) {
 
     withEnv(["PATH+EXTRA=/usr/bin:/usr/local/bin"]) {
         withCredentials([string(credentialsId: 's3_resource_bucket', variable: 'S3_BUCKET')]) {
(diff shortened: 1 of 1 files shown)


justice-adams-apple:
cc @azharudd

shahmishal (approved):

PR #678, opened 2025-12-12
[HIP] Add an --offload-arch option to hip-tpl.py

jmmartinez:
We add a `--offload-arch=...` option, that we can use to switch the target architecture. This can be a string such as `'gfx{version}'`, or `'amdgcnspirv'`.

This option defaults to 'gfx90a' such that if it is not specified, we keep the old behavior (`'gfx90a'` was hardcoded).

If we pass an incorrect option we get:
```bash
$ ./hip-tpl.py --offload-arch=aaa
usage: hip-tpl.py [-h] [--offload-arch OFFLOAD_ARCH]
hip-tpl.py: error: argument --offload-arch: invalid amdgpu_target value: 'aaa'
```

And for the help:

```bash
$ ./hip-tpl.py --help
usage: hip-tpl.py [-h] [--offload-arch OFFLOAD_ARCH]

options:
  -h, --help            show this help message and exit
  --offload-arch OFFLOAD_ARCH
                        Offload architecture to be forwarded to AMDGPU_ARCHS during the test suite configuration. e.g. gfx90a, gfx1010, amdgcnspirv
```

### DIFF
--- a/zorg/buildbot/builders/annotated/hip-tpl.py
+++ b/zorg/buildbot/builders/annotated/hip-tpl.py
@@ -17,12 +17,19 @@ def amdgpu_target(target):
 
 def main(argv):
     parser = argparse.ArgumentParser(prog=os.path.basename(__file__))
-    parser.add_argument("--offload-arch", default="gfx90a", type=amdgpu_target,
-                        help="Offload architecture to be forwarded to AMDGPU_ARCHS "
-                             "during the test suite configuration. "
-                             "e.g. gfx90a, gfx1010, amdgcnspirv")
-    parsed_args = parser.parse_args(argv[1:])
-    offload_arch = parsed_args.offload_arch
+    parser.add_argument("--offload-arch", action='append', type=amdgpu_target,
+                        help="Offload architectures to be forwarded to AMDGPU_ARCHS "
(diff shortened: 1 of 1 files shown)


zorg/buildbot/builders/annotated/hip-tpl.py
```diff
@@ -2,15 +2,28 @@
...
 import tempfile
 from contextlib import contextmanager
 
+def amdgpu_target(target):
+    if re.match("^(gfx[0-9a-f]+|amdgcnspirv)$", target):
+        return target
+    raise ValueError
 
 def main(argv):
+    parser = argparse.ArgumentParser(prog=os.path.basename(__file__))
```
jplehr, line 19: Would this also work with a combination, such as `gfx1030;gfx90a;gfx942` or so?

jmmartinez: > Would this also work with a combination, such as `gfx1030;gfx90a;gfx942` or so?

At the moment, no. But I can do that, it's not hard.

Should I add it?

jmmartinez: I've did it. Instead of passing a ";" separated string, I've made it take multiple `--offload-arch` parameters.

PR #680, opened 2025-12-15
[HIP] Add clang-hip-spirv builder

jmmartinez:
This is pretty much a "I have no idea of what I'm doing" kind of patch.
But I prefer to discuss around some code.


jplehr:
Thank you @jmmartinez 

This is also a bit more things to do on our end. So we will need to wait with this patch until we have a machine provisioned and a worker registered. Once those things have been done, we can go ahead with this patch.

PR #686, opened 2025-12-23
 Fix : Add timeout parameters to all HTTP requests

justynigam:
## Summary
Added timeout parameters to all HTTP requests to prevent indefinite hangs.

## Changes
- 5 files modified
- Added timeouts: 30s-300s depending on use case
- All tests passed

## Files
- submit-debuginfo-statistics-to-lnt.py (30s)
- submit-lldb-statistics-to-lnt.py (30s)
- process_llvm_commits.py (30s)
- build.py (60s)
- gcs.py (300s)

## Testing
✅ Syntax validation passed
✅ Timeout parameters verified
✅ Behavior testing passed

### DIFF
--- a/llvmbisect/llvmlab/gcs.py
+++ b/llvmbisect/llvmlab/gcs.py
@@ -73,7 +73,7 @@ CHUNK_SIZE = 5124288
 
 def get_compiler(url, filename):
     """Get the compiler at the url, and save to filename."""
-    r = HTTP_CLIENT.get(url)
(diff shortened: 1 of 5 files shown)


boomanaiden154 (changes requested):
Is this the result of timeouts actually occurring on the zorg infrastructure?

This does not look like it has been adequately tested, and in absence of actual timesout being observed, I'm hesitant we should move forward with this.

PR #687, opened 2025-12-26
[CI] Add flaky tests endpoint to premerge advisor

boomanaiden154:
The premerge advisor collects a large amount of information on test
executions across two LLVM CI systems. It would be good to expose the
information we have about what tests are flaky so they can be fixed.
This PR starts an initial pass at that, implementing the underlying DB
query/data processing using a simple heuristic, namely that the tests
has failed more than ten times across more than ten commits.

### DIFF
--- a/premerge/advisor/advisor_lib.py
+++ b/premerge/advisor/advisor_lib.py
@@ -32,6 +32,13 @@ class TestExplanationRequest(TypedDict):
     platform: str
 
 
+class FlakyTestInfo(TypedDict):
+    test_name: str
+    first_failed_index: str
+    last_failed_index: str
+    failure_range_commit_count: int
+
+
 _TABLE_SCHEMAS = {
     "failures": "CREATE TABLE failures(source_type, base_commit_sha, commit_index, source_id, test_file, failure_message, platform)",
     "commits": "CREATE TABLE commits(commit_sha, commit_index)",
@@ -161,7 +168,7 @@ def _try_explain_flaky_failure(
       db_connection: The database connection.
       test_failure: The test failure to try and explain.
       platform: The platform the test failed on.
-    
+
     Returns:
       Either None, if the test could not be explained as flaky, or a
       FailureExplanation object explaining the test failure.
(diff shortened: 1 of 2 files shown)


premerge/advisor/advisor_lib.py
```diff
@@ -222,3 +229,49 @@ def explain_failures(
...
+def get_flaky_tests(
+    db_connection: sqlite3.Connection,
+) -> list[FlakyTestInfo]:
+    possibly_flaky_tests = db_connection.execute(
+        "SELECT test_file, commit_index FROM failures where test_file "
+        "IN (SELECT test_file FROM failures GROUP BY test_file "
+        "HAVING COUNT(test_file) > 10)"
+    ).fetchall()
+    flaky_test_info: dict[str, FlakyTestInfo] = {}
+    flaky_test_count: dict[str, int] = {}
```
dwblaikie: Not sure why this is stored separately, rather than in the `FlakyTestInfo`?

dwblaikie, line 243: Seems like this maybe could be included in FlakyTestInfo?

premerge/advisor/advisor_lib_test.py
```diff
@@ -400,3 +400,48 @@ def test_no_explain_flaky_small_range(self):
...
+                (
+                    "postcommit",
+                    str(i * 10),
+                    i * 10,
+                    str(i * 10),
+                    "flaky_failing.ll",
+                    "test that is flaky",
+                    "linux-x86_64",
+                )
+            )
```
dwblaikie, line 429: So if I'm following this correctly you end up with [1,14] `consistently_failing.ll` (and since it's consistently failing over that range, it's not considered flaky - because it fails continuously and doesn't oscillate back and forth) then you have [10, 20, ... 140] are "flaky_failing.ll", resulting in 14 failures in 130 commit range, which is classified as flaky, since it oscillates.

A comment or two might be nice - and some testing closer to the boundary conditions I think might be clearer - a run of 9 failures, a pass, and another failure, compared to a run of 10 failures. (& maybe easier to read written out flat without loops, etc - copy paste isn't /so/ bad in a test case - but I do appreciate there's several fields here, and it'd only be especially legible if it used longer lines so each record was on a single line and you could scan down to see what's different, etc)

PR #691, opened 2025-12-27
[CI] Wire up premerge advisor debug mode

boomanaiden154:
Add a basic integration to ensure we are creating output and wire
everything up in server.py so we actually emit logs into an accessible
place.

### DIFF
--- a/premerge/advisor/advisor.py
+++ b/premerge/advisor/advisor.py
@@ -39,16 +39,12 @@ def explain():
     )
 
 
-<<<<<<< HEAD
 @advisor_blueprint.route("/flaky_tests")
 def flaky_tests():
     return advisor_lib.get_flaky_tests(_get_db())
 
 
-def create_app(db_path: str, repository_path: str):
-=======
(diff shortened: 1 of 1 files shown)


premerge/advisor/advisor.py
```diff
@@ -32,21 +32,29 @@ def upload():
...
+<<<<<<< HEAD
 @advisor_blueprint.route("/flaky_tests")
 def flaky_tests():
     return advisor_lib.get_flaky_tests(_get_db())
 
 
 def create_app(db_path: str, repository_path: str):
+=======
+def create_app(db_path: str, repository_path: str, debug_folder: str):
+>>>>>>> 932f0fe5 ([CI] Add debug logging for premerge advisor explanations)
```
dwblaikie: Remaining merge conflict?

boomanaiden154: Yep. Looks like I ran `spr` too eagerly. Should be fixed now.

PR #710, opened 2026-01-16
lldb-cmake-sanitized: update debugserver path

ndrewh:
debugserver now lives in stage2

rdar://164487647


ndrewh:
Merging to attempt to unblock lldb-cmake-sanitized

PR #712, opened 2026-01-16
lldb-cmake-sanitized should not be incremental

ndrewh:
The existing script says:

```
# Non-incremental, so always delete just in case.
rm -rf clang-build clang-install host-compiler *.tar.gz
```

But clang-build is not the build directory, `lldb-build` is. I noticed some strange behavior where the stage2 build wasn't being re-configured, and I think it's because of the incremental build.

https://green.lab.llvm.org/job/llvm.org/view/LLDB/job/lldb-cmake-sanitized/3587

Since this is now a bootstrapping build, we should not try to do an incremental build.


ndrewh:
Merging to try to unblock the lldb-cmake-sanitized job

PR #714, opened 2026-01-16
[AMDGPU][HIP] Listen to more schedulers

jplehr:
The CMake cache file that this bot uses enables a few more projects. Since the bot does collapse requests anyway, it won't add a lot to the turnaround time to add these schedulers.


Kewen12 (approved):
lgtm!

PR #720, opened 2026-01-28
Fixed the exception twisted.web._newclient.RequestGenerationFailed

slydiman:
Added @defer.inlineCallbacks before def is_wrong_issue to make twisted happy.

And never use `debug=True` in GitHubStatusPush and inherited classes.
It will cause unexpected exceptions inside HTTP client service.


gkistanova (approved):
LGTM.

PR #724, opened 2026-02-03
Add google-integrate buildbot init scripts

boomanaiden154:
These are primarily intended to service the
release-noassertions-warnings configurations as of now. These scripts will run as startup scripts for the GCP instances, run some setup, and then connect to the (staging currently) buildmaster and start servicing jobs.

### DIFF
--- a/buildbot/google/scripts/release-noassertions-warnings.sh
+++ b/buildbot/google/scripts/release-noassertions-warnings.sh
@@ -1,0 +1,46 @@
+set -ex
+
+apt-get update
+apt-get install -y python3 python3-pip cmake ninja-build git ccache lsb-release wget software-properties-common gnupg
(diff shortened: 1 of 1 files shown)


buildbot/google/scripts/release-noassertions-warnings.sh
```diff
@@ -0,0 +1,46 @@
+set -ex
```
cmtice, line 1: Shouldn't the first line of a *.sh script be a "#!" line?

boomanaiden154: It's not needed given the startup command will look something like
```
curl <url> | bash -
```
so we don't need the shebang given we're explicitly executing with bash. It wouldn't hurt, but it's also not canonical for the other buildbot scripts we have (e.g., profcheck, mlgo).

PR #732, opened 2026-02-09
[BazelBot] Add Dockerfile

boomanaiden154:
This patch adds the initial Dockerfile. This is a direct copy of the internal
version (with internal bug numbers removed). Cleanup will happen in future
PRs in llvm-zorg so it is visible in the git blame/history.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2200,6 +2200,8 @@ all += [
                         "-DCLANG_DEFAULT_LINKER=lld",
                         "-DLIBOMPTARGET_PLUGINS_TO_BUILD=level_zero",
                         "-DLIBOMPTARGET_ENABLE_DEBUG=ON",
+                        "-DLLVM_RUNTIME_TARGETS=default;spirv64-intel",
+                        "-DRUNTIMES_spirv64-intel_LLVM_ENABLE_RUNTIMES=openmp",
                     ])},
 
 


google-bazel-bot/docker/Dockerfile
```diff
@@ -0,0 +1,120 @@
+FROM ubuntu:latest
+
+# Make sure apt-get doesn't prompt for stuff like our time zone, etc.
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN echo 'intall packages'; \
```
cmtice, line 6: typo: 'intall' => 'install'

boomanaiden154: This is a direct copy+paste (minus removing internal bug references) from the internal version. I was planning on doing a thorough cleanup soon as there are a bunch of other things that need to be fixed too.

Given that I think it makes sense to postpone fixes like this, although if we want to fix ultra trivial issues like this now, I could be persuaded.

google-bazel-bot/docker/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+RUN curl -o sccache.tar.gz -L https://github.com/mozilla/sccache/releases/download/v0.8.1/sccache-v0.8.1-x86_64-unknown-linux-musl.tar.gz \
+    && echo "e0ee621fb16b6940666cd770b091c62cadafd3e062dd12e3a49d9caaff3b795f  sccache.tar.gz" | shasum -a 256 -c \
+    && tar xzf ./sccache.tar.gz \
+    && mv sccache-v0.8.1-x86_64-unknown-linux-musl/sccache /usr/bin \
+    && chown root:root /usr/bin/sccache \
+    && ls -la /usr/bin/sccache \
+    && sccache --version
+
+# LLVM must be installed after prerequsite packages.
+ENV LLVM_VERSION=18
```
cmtice, line 54: Is that the right LLVM version? It seems a bit old now (to me)...

google-bazel-bot/docker/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+  tar xzf /tmp/bant.tar.gz -C /tmp/; \
+  rm /tmp/bant.tar.gz; \
+  cp /tmp/bant-${BANT_VERSION}-linux-static-x86_64/bin/bant /usr/local/bin/;
+
+RUN echo 'install buildkite' ;\
+    apt-get install -y apt-transport-https gnupg;\
+    sh -c 'echo deb https://apt.buildkite.com/buildkite-agent stable main > /etc/apt/sources.list.d/buildkite-agent.list' ;\
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198 ;\
+    apt-get update ;\
+    apt-get install -y buildkite-agent tini gosu; \
```
cmtice, line 109: tini & gosu were apt-get installed back on line 10. Do we need them again here?

google-bazel-bot/docker/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+export SCCACHE_DIR="$BUILDKITE_BUILD_PATH/sccache"
+export SCCACHE_IDLE_TIMEOUT="0"
+rm -rf "$SCCACHE_DIR"
+mkdir -p "${SCCACHE_DIR}"
+chown -R ${USER}:${USER} "${SCCACHE_DIR}"
+chmod oug+rw "${SCCACHE_DIR}"
+gosu "$USER" bash -c 'SCCACHE_DIR="${SCCACHE_DIR}" SCCACHE_IDLE_TIMEOUT=0 SCCACHE_CACHE_SIZE=20G sccache --start-server'
+
+# /mnt/ssh should contain known_hosts, id_rsa and id_rsa.pub .
+mkdir -p /var/lib/buildkite-agent/.ssh
```
cmtice, line 25: Why not use ${BUILDKITE_BUILD_PATH} here and in the lines below?

google-bazel-bot/docker/pre-checkout
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# The `pre-checkout` hook will run just before your pipelines source code is
+# checked out from your SCM provider
+
+set -eu
+set -o pipefail
+
+gc_counter="$BUILDKITE_BUILD_PATH/git_gc_counter"
```
cmtice, line 9: What if $BUILDKITE_BUILD_PATH hasn't been defined yet?

google-bazel-bot/docker/pre-checkout
```diff
@@ -0,0 +1,36 @@
...
+echo "BUILDKITE_REPO: $BUILDKITE_REPO"
+if [ -d "$BUILDKITE_BUILD_CHECKOUT_PATH" ]; then
```
cmtice, line 11: I don't see where this env var gets defined?

cmtice (approved):

PR #735, opened 2026-02-09
Fix build artifact names

justice-adams-apple:
* Prepare the build.py script for use in multi branch pipelines by formatting JOB_NAMES which include encoded slashes
* Fix toolchain artifact names to remove build specific metadata, this way the artifact manager script can download a specific build of a toolchain based on it's git SHA/distance and does not need to know arbitrary build information. This will allow the script to reuse artifacts from bisection specific jobs


justice-adams-apple:
Do not merge this until the new pipelines are constructed to prevent breaking current CI functionality

PR #746, opened 2026-02-16
Add a new clang-x64-windows-msvc-forward-slashes builder.

matts1:
This builder is designed to be run to ensure that clang passes tests when -DLLVM_WINDOWS_PREFER_FORWARD_SLASH is enabled.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -659,6 +659,17 @@ all = [
                     script="clang-windows.py",
                     depends_on_projects=['llvm', 'clang', 'lld', 'debuginfo-tests'])},
 
+    {'name' : 'clang-x64-windows-msvc-forward-slashes',
+    'tags'  : ["clang"],
+    'workernames' : ['windows-gcebot2'],
+    'builddir': 'clang-x64-windows-msvc-forward-slashes',
+    'factory' : AnnotatedBuilder.getAnnotatedBuildFactory(
+                    script="clang-windows.py",
+                    extra_args=[
+                        '--extra_cmake_arg=-DLLVM_WINDOWS_PREFER_FORWARD_SLASH=ON',
+                    ],
+                    depends_on_projects=['llvm', 'clang', 'lld', 'debuginfo-tests'])},
+
     {'name' : "clang-m68k-linux",
     'tags'  : ["clang"],
     'workernames' : ["debian-akiko-m68k"],
--- a/zorg/buildbot/builders/annotated/clang-windows.py
+++ b/zorg/buildbot/builders/annotated/clang-windows.py
@@ -6,6 +6,7 @@ import annotated_builder
 
 def main(argv):
     ap = annotated_builder.get_argument_parser()
+    ap.add_argument('--extra_cmake_arg', nargs='*', default=[])
     args = ap.parse_args(argv[1:])
 
     # TODO: Add back debuginfo-tests once it works.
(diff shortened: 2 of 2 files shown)


buildbot/osuosl/master/config/builders.py
```diff
@@ -659,6 +659,17 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
                     script="clang-windows.py",
                     depends_on_projects=['llvm', 'clang', 'lld', 'debuginfo-tests'])},
 
+    {'name' : 'clang-x64-windows-msvc-forward-slashes',
+    'tags'  : ["clang"],
+    'workernames' : ['windows-gcebot2'],
```
matts1, line 664: This reuses an existing worker. I assumed this was fine since the intent was to run this builder infrequently. Please tell me if this isn't the case.

buildbot/osuosl/master/config/builders.py
```diff
@@ -659,6 +659,17 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
                     script="clang-windows.py",
                     depends_on_projects=['llvm', 'clang', 'lld', 'debuginfo-tests'])},
 
+    {'name' : 'clang-x64-windows-msvc-forward-slashes',
```
matts1, line 662: The consensus in https://github.com/llvm/llvm-project/pull/179865 was that this bot was important enough to run, but not important enough to spend CI resources to run as a pre-merge check.

I have no idea if this is even possible with buildbot, but my ideal scenario would be:
* Run it on a schedule (eg. 1/day)
* If a failure occurs, ideally it could bisect the failure to the failing commit and automatically comment on the commit saying that it broke some postsubmit tests.

What would be the recommended run configuration, and where / how can we configure it?

matts1, line 662: I assume we want to start with a staging builder. Is that as simple as adding a suffix "-staging"? Or is there no way to add a staging builder, only a staging worker?

buildbot/osuosl/master/config/builders.py
boomanaiden154: Builders don't run infrequently. They run as often as possible, up to doing a run on every commit. The original suggestion in the PR was to update the configuration of one of these builders, not add a new builder (from what I understand).

buildbot/osuosl/master/config/builders.py
boomanaiden154: The configuration is the same between the staging and production buildmasters. There are no naming differences for the builders. It just depends upon which buildmaster you connect your worker to (staging vs production).

buildbot/osuosl/master/config/builders.py
matts1: From what I can tell, there are mostly only two options for builders to reuse. The clang x64 windows builder, and the clang arm64 windows builder. We could modify an existing builder, though there'd be concerns that the builder would immediately start failing because the tests are not yet passing.

How would people feel about using the clang arm64 windows builder? We appear to have two of them - a 1 stage, and a 2 stage?

boomanaiden154: You'd have to ping one of the owners of that bot. Ideally the configuration should be green before switching a buldbot over to it.

matts1:
@gkistanova Could you PTAL, we have just finished fixing the last of the tests when it's enabled, and we'd like to prevent future breakages.

boomanaiden154 (commented):
Why do we need a new build instead of updating an existing configuration?

PR #748, opened 2026-02-17
[BazelBot] Clean up Dockerfile

boomanaiden154:
This patch cleans up the Dockerfile a bit:
1. We can disable a security mitigation now that we drop the package.
2. Install CMake from the repositories now that it is new enough.
3. Drop sccache given we do not need it for the bazel image.
4. Drop bazel bot specific packages from the base Dockerfile (will be added
   in a stage 2 build later).
5. Some other misc cleanup.

### DIFF
--- a/google-bazel-bot/docker/Dockerfile
+++ b/google-bazel-bot/docker/Dockerfile
@@ -57,7 +57,7 @@ RUN wget --no-verbose -O /usr/bin/bazelisk https://github.com/bazelbuild/bazelis
   chmod +x /usr/bin/bazelisk; \
   bazelisk --version
 
-# Build and install buildifier
+# Build and install buildifier and buildozer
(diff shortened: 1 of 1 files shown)


google-bazel-bot/docker/Dockerfile
```diff
@@ -85,7 +57,7 @@ RUN wget --no-verbose -O /usr/bin/bazelisk https://github.com/bazelbuild/bazelis
   chmod +x /usr/bin/bazelisk; \
   bazelisk --version
 
-# Build and install buildifier and buildozer.
+# Build and install buildifier
```
cmtice: You removed "buildozer" from the comment, but you haven't removed it from any of the build statements below. Why?

boomanaiden154: Fixed. Meant to add it back in when readding buildozer while looking at moving it to the stage2 build, but decided against it.

PR #755, opened 2026-02-18
[BazelBot] Add a service account for GCS authentication

boomanaiden154:
We need permissions to access the cache GCS bucket. Add a service account on
the GCP/k8s side so that we can allow the pod to access the bucket and avoid
403 errors during builds.

### DIFF
--- a/google-bazel-bot/main.tf
+++ b/google-bazel-bot/main.tf
@@ -66,6 +66,9 @@ resource "google_container_node_pool" "bazel_ci" {
 
   node_config {
     machine_type = "n2-standard-64"
+    workload_metadata_config {
+      mode = "GKE_METADATA"
+    }
   }
 }
 


google-bazel-bot/bazel-buildkite.yaml
```diff
@@ -13,6 +13,7 @@ spec:
       labels:
         app: bazel-buildkite
     spec:
+      serviceAccountName: bazel-cache-ksa
```
cmtice, line 16: Is this a typo? Should it be "bazel-cache-gsa"?

google-bazel-bot/main.tf
```diff
@@ -65,6 +69,29 @@ resource "google_container_node_pool" "bazel_ci" {
   }
 }
 
+resource "google_service_account" "bazel_cache_gsa" {
```
cmtice, line 75: The name above (in the yaml file) used hyphens, this uses underscores. Is that ok? Are they supposed to be the same?

google-bazel-bot/bazel-buildkite.yaml
boomanaiden154: No. This is the kubernetes service account (KSA) which is distinct from the google service account (GSA). There are definitions for both in the terraform.

google-bazel-bot/main.tf
boomanaiden154: Yes. The name used inside of k8s is hyphenated and specified on line 131 in `main.tf`.

PR #768, opened 2026-02-24
[BazelBot] Declare Bazel Cache Bucket in TF

boomanaiden154:
This declaratively creates the GCS bucket which makes it clear what is actually set up, and also makes maintenance easier in the future.

### DIFF
--- a/google-bazel-bot/main.tf
+++ b/google-bazel-bot/main.tf
@@ -95,6 +95,9 @@ resource "google_storage_bucket" "bazel_cache_bucket" {
   uniform_bucket_level_access = true
   public_access_prevention    = "enforced"
 
+  # Disable soft deletion. Soft deletion keeps the file around for the
+  # specified duration (or deletes immediately if set to zero) after it is
+  # requested to be deleted.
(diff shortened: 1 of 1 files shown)


google-bazel-bot/main.tf
```diff
@@ -88,14 +88,29 @@ resource "google_project_iam_binding" "vertex_ai_user_binding" {
...
-  name = "llvm-bazel-cache"
+resource "google_storage_bucket" "bazel_cache_bucket" {
+  name     = "llvm-bazel-cache"
+  location = "us-central1-c"
+
+  uniform_bucket_level_access = true
+  public_access_prevention    = "enforced"
+
+  soft_delete_policy {
+    retention_duration_seconds = 0
```
nigham, line 102: Presumably this means "never delete"? Could you leave a comment?

boomanaiden154: This actually means "delete immediately", as in don't soft delete (where a deletion actually leaves the file around for a while), but just fully delete. Added a comment.

PR #777, opened 2026-03-02
[Premerge] Remove libc++ runner command configuration

boomanaiden154:
Now that the release/next/current sets are all consistent, we can just hard code this value.

### DIFF
--- a/premerge/libcxx_runners_values.yaml
+++ b/premerge/libcxx_runners_values.yaml
@@ -22,7 +22,7 @@ template:
     containers:
     - name: runner
       image: ${ runner_image }
-      command: ["/home/agha/actions-runner/run.sh"]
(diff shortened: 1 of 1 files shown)


premerge/libcxx_runners_values.yaml
```diff
@@ -22,7 +22,7 @@ template:
     containers:
     - name: runner
       image: ${ runner_image }
-      command: ["${ command }"]
+      command: ["/home/agha/actions-runner/run.sh"]
```
cmtice: The commands you removed were all '/home/gha/...' this one is '/home/agha/...'

Is the 'a' in front of 'gha' a typo?

boomanaiden154: Yes. Thanks for catching that.

PR #781, opened 2026-03-04
[BazelBot] Use Github App for Authentication

boomanaiden154:
This patch makes it so that bazelbot uses a Github App for
authentication instead of a user token. This is best practice for
authenticating with Github and ensures that there is not a single person
responsible for the token.

### DIFF
--- a/google-bazel-bot/main.tf
+++ b/google-bazel-bot/main.tf
@@ -140,7 +140,7 @@ data "google_secret_manager_secret_version" "buildkite_agent_token" {
 }
 
 data "google_secret_manager_secret_version" "github_app_id" {
-  secret = "github-api-id"
+  secret = "github-app-id"
 }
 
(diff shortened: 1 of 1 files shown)


google-bazel-bot/main.tf
```diff
@@ -139,8 +139,12 @@ data "google_secret_manager_secret_version" "buildkite_agent_token" {
   secret = "buildkite-agent-token"
 }
 
-data "google_secret_manager_secret_version" "github_api_token" {
-  secret = "github-api-token"
+data "google_secret_manager_secret_version" "github_app_id" {
+  secret = "github-api-id"
```
cmtice: Question: Is the supposed to be 'github-api-id' or 'github-app-id'?

boomanaiden154: Meant to be `app`. Fixed.

PR #783, opened 2026-03-04
[BazelBot] Fix import sorting/style

boomanaiden154:
Makes us more compliant/consistent with
https://google.github.io/styleguide/pyguide.html which I think is
generally reasonable to follow in LLVM given the absence of LLVM
specific python standards.


google-bazel-bot/bazel-bot/bazel_agent.py
```diff
@@ -1,25 +1,24 @@
...
+from google.adk.agents.invocation_context import InvocationContext
+from google.adk.events import Event
 from google.adk.models.google_llm import _ResourceExhaustedError
 from google.adk.runners import Runner
 from google.adk.sessions import InMemorySessionService
-from google.adk.agents.invocation_context import InvocationContext
-from google.adk.events import Event
 from google.genai.types import Content, Part
 
+import tools
```
cmtice, line 19: Why are these imports here rather than up around line 8?

google-bazel-bot/bazel-bot/tools.py
```diff
@@ -1,18 +1,19 @@
-import os
 import json
+import os
+import pathlib
+from typing import Any
+
 import requests
```
cmtice, line 6: Shouldn't line 6 actually come between lines 3 & 4?

google-bazel-bot/bazel-bot/utils.py
```diff
@@ -1,28 +1,28 @@
...
-from enum import Enum
-from typing import List, Optional, Any
+import subprocess
+import time
+from collections.abc import Mapping, Sequence
+from typing import Any
+
+import git
+import github
+import requests
```
cmtice, line 14: Shouldn't  lines 12-14 be inserted alphabetically in lines 2-10?

google-bazel-bot/bazel-bot/bazelbot_server_test.py
```diff
@@ -1,31 +1,23 @@
...
-    BuildkiteBuildProcessor,
-)
-from bazelbot_server import BazelRepairBot
-from bazel_agent import AgentErrors
+import unittest
+from unittest import mock
+
+import bazel_agent
+import bazelbot_server
+import utils
```
cmtice, line 7: Imports should all be in alphabetical order?

google-bazel-bot/bazel-bot/bazel_agent.py
boomanaiden154: Here and below:

Import ordering is typically done as follows:
1. Imports of builtin python modules
2. line break
3. Imports of third party installed python modules
4. line break
5. Imports of modules from within the current project.

I've just used `isort` to sort these, which does this mechanically.

PR #811, opened 2026-03-16
[BazelBot] Add exponential backoff to LLM queries

boomanaiden154:
This seems to get rid of the HTTP 429 errors that we were seeing. This
patch does the following:
1. Moves from regular to priority quota. This is more costly, but likely
   necessary for our workload given the usage is incredibly bursty (lots
   of usage to fix a build break, otherwise none). Without this, we
   continue to get 429s until we run out of retrys.
2. Adds a number of retrys so that we do not immediately bail if we hit
   a 429.

### DIFF
--- a/google-bazel-bot/bazel-bot/bazel_agent.py
+++ b/google-bazel-bot/bazel-bot/bazel_agent.py
@@ -28,6 +28,7 @@ parser.add_argument(
 )
 
 MAX_AGENT_ITERATIONS = 5
+MAX_LLM_QUERY_RETRIES = 20
 USER_ID = "default_user"
 APP_NAME = "bazel_fixer"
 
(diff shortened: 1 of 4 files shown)


google-bazel-bot/bazel-bot/bazel_agent.py
```diff
@@ -81,6 +81,16 @@
         tools.get_diff,
         tools.directory_structure,
     ],
+    generate_content_config=types.GenerateContentConfig(
+        http_options=types.HttpOptions(
+            retry_options=types.HttpRetryOptions(initial_delay=1, attempts=20),
```
nigham: Optional: making the #attempts a flag might help with future testing?

boomanaiden154: The plumbing for a flag is probably overkill. A module level constant does make sense though.

PR #813, opened 2026-03-16
[BazelBot] Ensure bant is installed

boomanaiden154:
This was accidentally removed in
a4467ea49f36195a6a33f781fb17d689d9a7eeef. The bot does not hard fail on
missing bant, instead it just goes directly to the LLM.

### DIFF
--- a/buildbot/osuosl/master/config/status.py
+++ b/buildbot/osuosl/master/config/status.py
@@ -498,11 +498,9 @@ def getReporters():
                         "openmp-offload-amdgpu-runtime",
                         "openmp-offload-amdgpu-runtime-2",
                         "openmp-offload-libc-amdgpu-runtime",
-                        "openmp-offload-sles-build-only",
                         "amdgpu-offload-ubuntu-22-cmake-build-only",
-                        "amdgpu-offload-rhel-8-cmake-build-only",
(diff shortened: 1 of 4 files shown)


google-bazel-bot/docker/Dockerfile
```diff
@@ -88,6 +88,13 @@ CMD ["gosu", "buildkite-agent", "buildkite-agent", "start", "--no-color", "--tag
 
 FROM bazel-base as bazel-fixer-bot
 
+# Build and install bant.build to automatically fix issues by bazelbot.
+ENV BANT_VERSION=v0.2.3
+RUN wget --no-verbose -O /tmp/bant.tar.gz https://github.com/hzeller/bant/releases/download/${BANT_VERSION}/bant-${BANT_VERSION}-linux-static-x86_64.tar.gz; \
```
nigham, line 93: I assume this is not available via apt-get etc? Is there a plan to support that? (asking because hardcoding a version / arch etc seems brittle).

boomanaiden154: Yeah, it's not available with `apt-get`.

I would say this approach is less fragile than just pulling stuff in from `apt-get`. This ensures we do not get upgrades without our knowledge. This is also relatively standard practice for building containers.

PR #816, opened 2026-03-19
[BazelBot] Consolidate validation logic

### LINKED ISSUE #795: Bazel Fixer Bot not running buildifier on .bzl files
See details in https://github.com/llvm/llvm-project/pull/185223

### LINKED ISSUE #802: Bazel fixer bot will create empty PRs
e.g., https://github.com/llvm/llvm-project/pull/186691

This was likely a transient failure (buildkite doesn't show anything), so an empty fix works. We should prevent creating PRs for empty commits.

boomanaiden154:
This patch makes it so that we reuse the validation logic for bant for the LLM based fix loop as well. This fixes at least two problems, namely running buildifier after the build and pushing empty commits. This does take away the ability to see attempted fixes that do not pass the build, but I have never seen one of these pop up, and their utility is questionable.

Fixes #802.
Fixes #795.

### DIFF
--- a/google-bazel-bot/bazel-bot/bazelbot_server.py
+++ b/google-bazel-bot/bazel-bot/bazelbot_server.py
@@ -182,22 +182,18 @@ class BazelRepairBot:
 
             attempt_num += 1
 
-        bazel_build = self.command_processor.run_bazel_build()
-        commit_msg = f"Fix Bazel build for {build_data.commit[:7]}"
-        if not bazel_build.success:
-            logger.info("Local verification FAILED after running AI agent.")
-            logger.info(bazel_build.stderr)
-            commit_msg = f"[DO NOT MERGE] Attempted AI fix for {build_data.commit[:7]}"
+        if not self.validate_before_publishing():
+            logger.info("Validation before publishing failed.")
+            return False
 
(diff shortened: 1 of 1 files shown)


google-bazel-bot/bazel-bot/bazelbot_server.py
nigham, line 105: Should this just be renamed 'validate_build`?

The "before publishing" part is somewhat confusing, and it assumes usage of the function in a certain way.

nigham, line 116: Should we do this before running the bazel build?

boomanaiden154: We run it twice, so there are no correctness concerns. If buildifier behaves weirdly, it might be nice to be able to tease the failure modes apart using just the log information.

boomanaiden154: It does more than validate the build. We're also checking things like empty commits/formatting.

And I don't think there is going to be any usage of this function outside of fix loops calling it right before creating PRs (the publishing part).

nigham:
> Couple nits/questions but otherwise LGTM

Sorry didn't see that you'd replied. LGTM

PR #817, opened 2026-03-19
[BazelBot] Close existing PRs when pushing a fix

boomanaiden154:
Whenever we push a new fix, check if there are any open PRs also created by the bazelbot and close them. If we are creating a new fix, we are guaranteed to have a green->red transition after the previous fix was made.

This prevents the bot from cluttering up the PR list which isn't a huge deal, but is slightly annoying.

### DIFF
--- a/google-bazel-bot/bazel-bot/bazelbot_server_test.py
+++ b/google-bazel-bot/bazel-bot/bazelbot_server_test.py
@@ -80,7 +80,7 @@ class TestBazelBotServer(unittest.TestCase):
 
             @property
             def url(self) -> str:
-                return "testing"
+                return "PullRequestMockObjectURL"
 
         mock_prs = [MockPullRequest(), MockPullRequest()]
         repo.gh_pr_repo.get_issues = mock.MagicMock()


google-bazel-bot/bazel-bot/utils.py
```diff
@@ -242,6 +242,13 @@ def commit(self, message: str) -> None:
     def is_repo_dirty(self, untracked_files=False) -> bool:
         return self.repo.is_dirty(untracked_files=untracked_files)
 
+    def close_existing_prs(self) -> None:
```
nigham, line 245: I wonder if we might want this bot to work in parallel at some point and closing wholesale unmerged PRs may not be the right thing? (e.g. if there are multiple breakages being handled separately).

But OK to revisit that later even if you think it's useful.

google-bazel-bot/bazel-bot/bazelbot_server_test.py
```diff
@@ -70,11 +70,29 @@ def test_local_git_repo(
...
+        class MockPullRequest:
+            def __init__(self):
+                setattr(self, "state", "open")
+
+            def edit(self, state: str) -> None:
+                setattr(self, "state", state)
+
+            @property
+            def url(self) -> str:
+                return "testing"
```
nigham: s/testing/PullRequestMockObject?

google-bazel-bot/bazel-bot/utils.py
boomanaiden154: Yeah, I think this is something we revisit later if/when we need it.

PR #818, opened 2026-03-19
[BazelBot] Remove use of GITHUB_API_TOKEN

boomanaiden154:
This API token was only kept around so that the ADK could use it inside of a tool. We side step this problem by removing the agent from global state, doing some token plumbing, and making it so that we can create a diff tool with a specific token. We can use a single token per LLM invocation as they expire after an hour, which should be fine.

We need to be somewhat careful about how we construct the tool function. The Google ADK relies on the docstring to pass along to the agent, so we need to make sure that we pass along a real function and not something like a functools.partial.

### DIFF
--- a/google-bazel-bot/bazel-bot/tools.py
+++ b/google-bazel-bot/bazel-bot/tools.py
@@ -107,6 +107,20 @@ def search_and_replace(file_path: str, old_content: str, new_content: str) -> st
 
 
 def get_diff_tool(github_token: str):
+    """Returns a function containing a tool to get a diff.
+
+    Takes in as argument a github token to use when accessing the github API
+    and returns a function that can be used as a LLM tool for getting a diff
+    for a specific commit.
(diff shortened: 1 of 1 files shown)


google-bazel-bot/bazel-bot/tools.py
```diff
@@ -106,30 +106,29 @@ def search_and_replace(file_path: str, old_content: str, new_content: str) -> st
         return f"Error modifying file: {str(e)}"
 
 
-def get_diff(commit_sha: str) -> str:
-    """
-    Retrieves the git diff for a specific commit SHA using the GitHub API.
+def get_diff_tool(github_token: str):
```
nigham, line 109: Document this as well? (returns a function object..)

boomanaiden154: Done.

PR #827, opened 2026-03-31
Add additional CMake options to PS targeting bots

dyung:
We would like to upstream some of our downstream changes and use a few additional CMake configuration options so that our bot will test it and it can make sure it isn't broken by others. This change adds the following three options to our PS targeting bots and once they are committed we will work on upstreaming our downstream changes that make use of these options.

- CLANG_RESOURCE_DIR
- CLANG_VENDOR
- CLANG_VENDOR_UTI


gkistanova (approved):
LGTM.

PR #836, opened 2026-04-14
[sanitizers] Preparation for P3953R3: Rename `std::runtime_format`

frederick-vs-ja:
See
- https://wg21.link/p3953r3
- https://llvm.org/PR189657

Revert
- https://llvm.org/PR191912

It would make more sense to rename the test files to mention `dynamic_format`. For compatibility, this patch temporarily use both old and new names.


thurstond (approved):
Thanks!

PR #840, opened 2026-04-29
Remove worker no longer running

jpienaar:
This should no longer be needed.


gkistanova (approved):
Thanks for the patch, Jacques.

PR #842, opened 2026-05-01
[AIX] adjust thread numbers for aix builder

madanial0:
A bunch of builds fails due to current width of 80, reduce it to 75


gkistanova:
Thanks, Mark!

Are you confident that it will work reliably with 75 jobs?

madanial0:
I am fairly confident but if issues persist we can try lowering it further.

PR #848, opened 2026-05-14
[lldb][windows] Remove LLDB_USE_LLDB_SERVER

charles-zablit:
`LLDB_USE_LLDB_SERVER` has no effect when running tests, the environment variable is not herited by the `lldb.exe` subprocess when running tests: https://github.com/llvm/llvm-project/pull/197660.

Furthermore, `lldb-server` on Windows does not work well and 100+ tests in `check-lldb` are failing when it's used.

This patch removes this `LLDB_USE_LLDB_SERVER` before merging https://github.com/llvm/llvm-project/pull/197660, as it will enable the `lldb-server` tests, which will fail. It's not tested at the moment.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -1365,7 +1365,6 @@ all = [
                         "LLDB_TEST_USER_ARGS"           : "--skip-category=lldb-dap",
                     },
                     env = {
-                        'LLDB_USE_LLDB_SERVER' : "1",
                         'CCACHE_DIR'    : util.Interpolate("%(prop:builddir)s/ccache-db"),
                         # TMP/TEMP within the build dir (to utilize a ramdisk).
                         'TMP'           : util.Interpolate("%(prop:builddir)s/build"),


DavidSpickett (approved):
Depending on how intentional it was to set it to 1, the bot owner might be a bit surprised that it was not doing anything. Maybe they will be interested in your work on this topic. So wait for them to review this too.

That said, I agree with removing it for now given that it does nothing but mislead.

vvereschaka:
@charles-zablit ,
>Furthermore, lldb-server on Windows does not work well and 100+ tests in check-lldb are failing when it's used.

I don't see 100+ failed tests on the buildbot's builder (i.e https://lab.llvm.org/buildbot/#/builders/211). Are you sure you build and run the proper build configuration on your side? Have you tested these changes on x86 Windows host?
As far as I remember this env var is required to run the localhost lldb-server tests properly on Windows/x86_64.

@slydiman do you remember anything about this env var in order of the x86_64 windows builder?

charles-zablit:
> @charles-zablit ,
> 
> > Furthermore, lldb-server on Windows does not work well and 100+ tests in check-lldb are failing when it's used.
> 
> I don't see 100+ failed tests on the buildbot's builder (i.e [lab.llvm.org/buildbot#/builders/211](https://lab.llvm.org/buildbot/#/builders/211)). Are you sure you build and run the proper build configuration on your side? Have you tested these changes on x86 Windows host? As far as I remember this env var is required to run the localhost lldb-server tests properly on Windows/x86_64.
> 
> @slydiman do you remember anything about this env var in order of the x86_64 windows builder?

`LLDB_USE_LLDB_SERVER=1` never reaches the tests, so `lldb-server` is not used on the bots. This is due to the testing configuration not allowing the env variable to propagate. https://github.com/llvm/llvm-project/pull/197660 fixes this. Once it lands, with the current config, a lot of tests will fail.

slydiman:
If LLDB_USE_LLDB_SERVER is set to 1/on/yes/true the LLDB plugin `ProcessWindows` is not registered and lldb-server on localhost is used instead.

> LLDB_USE_LLDB_SERVER=1 never reaches the tests

The behavior may be different at least for API and shell tests.

I checked may scripts - I'm using `SET LLDB_USE_LLDB_SERVER=1` before `ninja check-lldb` in local tests on Windows x86_64 because this is necessary for a correct result.

slydiman (approved):
I have retested `check-lldb` on Windows x86_64 locally without LLDB_USE_LLDB_SERVER and got the same result.
So, I think we can remove it for now.

vvereschaka (approved):
Thank you @charles-zablit @slydiman

vvereschaka:
I'm going to merge this PR to catch today's configuration update the production master.

PR #854, opened 2026-05-19
[dep] Delete dep

boomanaiden154:
This code does not appear to be used anywhere and has not been touched in a long time, so delete it.


JDevlieghere (approved):

boomanaiden154:
Looks like this one is actually still used in zorg/jenkins/old_build.py, which looks used.

PR #856, opened 2026-05-28
Add check-clang-unit, check-llvm-unit with LLVM_WINDOWS_PREFER_FORWARD_SLASH enabled on llvm-clang-x86_64-win-fast bot

Jwata:
This PR enables running `check-clang-unit`, `check-llvm-unit` tests twice on `llvm-clang-x86_64-win-fast` bot: once with default settings (backslash path behavior), and once with the forward slash override enabled (`LLVM_WINDOWS_PREFER_FORWARD_SLASH=1`).

### Context & Rationale
There is no test coverage for LLVM_WINDOWS_PREFER_FORWARD_SLASH feature. Adding LLVM_WINDOWS_PREFER_FORWARD_SLASH enabled tests will improve the coverage, so that the change owners can notice test breakages.
It was also [discussed](https://discord.com/channels/636084430946959380/642340762620657675/1506736958611652658) to add tests with the flag to  an existing bot, rather than adding a new builder.

### Key Changes
*   `llvm-clang-x86_64-win-fast` will run the following tests:
    1.  `check-llvm-unit` (Default)
    2.  `check-clang-unit` (Default)
    3.  `check-llvm-unit-forward-slashes` (with `LLVM_WINDOWS_PREFER_FORWARD_SLASH=1` override)
    4.  `check-clang-unit-forward-slashes` (with `LLVM_WINDOWS_PREFER_FORWARD_SLASH=1` override)

* Introduces a new check type with `target`, `env_override`, `name_suffix` to support overriding an environment variable.

### Questions for reviewers
* Is there a better way than the `env_override` approach?
* Should `llvm-clang-x86_64-expensive-checks-win` also duplicate the checks, too?

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -92,7 +92,17 @@ all = [
                     clean=True,
                     checks=[
                         "check-llvm-unit",
-                        "check-clang-unit"
+                        "check-clang-unit",
+                        {
+                            'target': 'check-llvm-unit',
+                            'env_override': {'LLVM_WINDOWS_PREFER_FORWARD_SLASH': '1'},
+                            'name_suffix': '-forward-slashes'
+                        },
+                        {
+                            'target': 'check-clang-unit',
+                            'env_override': {'LLVM_WINDOWS_PREFER_FORWARD_SLASH': '1'},
+                            'name_suffix': '-forward-slashes'
+                        }
                     ],
                     extra_configure_args=[
                         "-DLLVM_CCACHE_BUILD=ON",
--- a/zorg/buildbot/builders/UnifiedTreeBuilder.py
+++ b/zorg/buildbot/builders/UnifiedTreeBuilder.py
@@ -13,6 +13,7 @@ from zorg.buildbot.conditions.FileConditions import FileDoesNotExist
 from zorg.buildbot.process.factory import LLVMBuildFactory
 
 import zorg.buildbot.builders.Util as builders_util
+from zorg.buildbot.process.properties import MergedEnv
 
 def getLLVMBuildFactoryAndPrepareForSourcecodeSteps(
            depends_on_projects = None,
@@ -235,13 +236,28 @@ def addNinjaSteps(
 
     if checks:
         for check in checks:
-            f.addStep(LitTestCommand(name=trunc50("test-%s-%s" % (step_name, check)),
-                                    command=['ninja', check],
+            check_target = check
+            check_env_override = None
+            check_suffix = ""
+
+            if isinstance(check, dict):
+                check_target = check.get('target')
+                check_env_override = check.get('env_override')
+                check_suffix = check.get('name_suffix', "")
+
+            check_name = check_target + check_suffix
+
+            step_env = check_env
+            if check_env_override:
+                step_env = MergedEnv(check_env, check_env_override)
+
+            f.addStep(LitTestCommand(name=trunc50("test-%s-%s" % (step_name, check_name)),
+                                    command=['ninja', check_target],
                                     description=[
                                     "Test", "just", "built", "components", "for",
-                                    check,
+                                    check_name,
                                     ],
-                                    env=check_env,
+                                    env=step_env,
                                     workdir=obj_dir,
                                     **kwargs # Pass through all the extra arguments.
                                     ))
--- a/zorg/buildbot/process/properties.py
+++ b/zorg/buildbot/process/properties.py
@@ -59,3 +59,26 @@ class InterpolateToPosixPath(Interpolate):
             pass
 
         return p
+
+
+@implementer(IRenderable)
+class MergedEnv(object):
+    def __init__(self, base_env, overrides):
+        self.base_env = base_env
+        self.overrides = overrides
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, build):
+        if hasattr(build, 'render'):
+            base_env_rendered = yield build.render(self.base_env)
+        elif hasattr(build, 'properties') and hasattr(build.properties, 'render'):
+            base_env_rendered = yield build.properties.render(self.base_env)
+        else:
+            base_env_rendered = {}
+
+        if not isinstance(base_env_rendered, dict):
+            base_env_rendered = {}
+
+        merged = base_env_rendered.copy()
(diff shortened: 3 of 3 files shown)


zmodem:
Can you provide a non-AI-generated description?

Jwata:
> Can you provide a non-AI-generated description?

Hmm, it might be too verbose. I made it concise. PTAL

Jwata:
@boomanaiden154 
Can you take a look or help find someone can review this PR?

buildbot/osuosl/master/config/builders.py
```diff
@@ -92,7 +92,17 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
                     clean=True,
                     checks=[
                         "check-llvm-unit",
-                        "check-clang-unit"
+                        "check-clang-unit",
+                        {
+                            'target': 'check-llvm-unit',
+                            'env_override': {'LLVM_WINDOWS_PREFER_FORWARD_SLASH': '1'},
```
boomanaiden154, line 98: Why exactly does this need to be passed as an env variable?

It seems like this should just be a CMake option. https://github.com/llvm/llvm-project/blob/56ccbc25315030088b04435e4828efbc22f0e926/llvm/CMakeLists.txt#L638

boomanaiden154:
> Can you take a look or help find someone can review this PR?

You need to get approval from the bot's owner.

buildbot/osuosl/master/config/builders.py
Jwata: This allows us to build only once and run tests twice: the first ones with default, and the second ones with LLVM_WINDOWS_PREFER_FORWARD_SLASH feature, rather than  building twice or duplicating the bot.

Jwata:
> > Can you take a look or help find someone can review this PR?
> 
> You need to get approval from the bot's owner.

I see. Thanks.

@gkistanova Could you take a look?

rnk (changes requested):
Instead of configuring the buildbot level to build the `check*unit*` targets twice in different environments, I would configure this at the lit test suite level. You can look at the sanitizer tests, which do some matrix testing (re-running the test suite in multiple configurations). They do this in a way that allows you to run all the tests in a single lit invocation, reducing bottlenecks, improving utilization, and reducing cycle time.

You could control that via a cmake option.

vvereschaka:
Thank you for PR @Jwata , but there are few moments:

* a goal of the `llvm-clang-x86_64-win-fast`  builder is a building of each commit for the main components -- LLVM/Clang -- as fast as it possible and adding two additional steps of the testing increases an average build time for 30-35%. It seems to much for me and I wouldn't like to slowdown *this* builder that much at least for now.

* actually, you don't need to run both of those checks on the same builder. There are few more Windows bots running the LLVM/Clang unit tests (check-clang/check-llvm or check-all) and they already testing the build config without `LLVM_WINDOWS_PREFER_FORWARD_SLASH` (*-win-x-* as example). You'll need just a builder that checks a build with `LLVM_WINDOWS_PREFER_FORWARD_SLASH=ON`. We can add `-DLLVM_WINDOWS_PREFER_FORWARD_SLASH=ON` as CMake parameter of the build configuration for this builder and check it here without any additional checks.

* I see a failed LLVM-Unit test when I built with `-DLLVM_WINDOWS_PREFER_FORWARD_SLASH=ON` on my local computer:
```
********************
Failed Tests (1):
  LLVM-Unit :: Support/./SupportTests.exe/FileSystemTest/CreateRelativeDirectorySymlink


Testing Time: 13.42s

Total Discovered Tests: 11096
  Skipped:   644 (5.80%)
  Passed : 10451 (94.19%)
  Failed :     1 (0.01%)
```
This test must be fixed before applying that changes on the buildbot configuration.

* Also, I don't see this failure when build without CMake's `LLVM_WINDOWS_PREFER_FORWARD_SLASH`, but run them with `set LLVM_WINDOWS_PREFER_FORWARD_SLASH=1` environment variable. Looks like the environment variable is not supported at all (or does not work as expected).

PS. You can find the owners of the builders on https://lab.llvm.org/buildbot/#/workers page at **Admin** column.

Jwata:
> Instead of configuring the buildbot level to build the `check*unit*` targets twice in different environments, I would configure this at the lit test suite level. You can look at the sanitizer tests, which do some matrix testing (re-running the test suite in multiple configurations). They do this in a way that allows you to run all the tests in a single lit invocation, reducing bottlenecks, improving utilization, and reducing cycle time.
> 
> You could control that via a cmake option.

@rnk 
Sorry for the late reply. I drafted https://github.com/llvm/llvm-project/pull/207194. But I'm not sure if that is aligned with what you proposed. It looks a bit complex for me. So wanted to check if that is acceptable.

Jwata:
@vvereschaka 

Thank you for your reply and sorry for the late reply...

>  goal of the llvm-clang-x86_64-win-fast builder is a building of each commit for the main components -- LLVM/Clang -- as fast as it possible

Sure. That's good to know.

> You'll need just a builder that checks a build with LLVM_WINDOWS_PREFER_FORWARD_SLASH=ON. We can add -DLLVM_WINDOWS_PREFER_FORWARD_SLASH=ON as CMake parameter of the build configuration for this builder and check it here without any additional checks.

It looks like each builder has a dedicated worker, IIUC. So do I understand correctly that you are suggesting to have a build to share some of the workers with an existing builder? You are the owner of many of  Windows `as-builder-*`.  Do you have any recommendation for the builder that shares the worker with the new builder?

>Failed Tests (1):
>  LLVM-Unit :: Support/./SupportTests.exe/FileSystemTest/CreateRelativeDirectorySymlink

Thank you for letting me know the regression. It should be fixed by https://github.com/llvm/llvm-project/pull/206665.

vvereschaka:
@Jwata 

> It looks like each builder has a dedicated worker,

Slightly differently, each worker (host) has assigned builder(s) (one or more). Each builder builds it's own build configuration (`llvm-clang-x86_64-win-fast` as example). I don't suggest to share the existing worker for a new builder, this is too resource expensive to create one more builder on the existing workers for just that kind of option. I just suggest to add `-DLLVM_WINDOWS_PREFER_FORWARD_SLASH=ON` to one of the existing build configurations. 

Also, you always can set up your own worker with a set of desired builders (and the proper build configurations) and attach it to the public buildbot.

PR #860, opened 2026-06-02
[Buildbot][AMDGPU] Onboard a new buildbot

Kewen12:
Onboard a new buildbot for AMDGPU.


jplehr (approved):

jplehr:
Do we want to update the notifier file already with this PR so we don't forget about it later?

Kewen12:
> Do we want to update the notifier file already with this PR so we don't forget about it later?

Good callout! Updated the notifier in the revision.

PR #861, opened 2026-06-03
[HIP] Rely on CMake cache for compiler-rt

jplehr:
The script enabled the compiler-rt explicitly. However, it was/is already enabled in the CMake cache file and our other builders rely on the CMake cache alone.

Remove it as an explicit setting in this bot.


jplehr:
This difference in cmake invocation may have been the culprit of the breakage reported in https://lab.llvm.org/buildbot/#/builders/230/builds/1653

jplehr:
I'm going to land this and see if it unblocks the buildbot.

PR #863, opened 2026-06-03
[BOLT][AArch64] Do not generate A53 veneers

paschalis-mpeis:
GCC generates erratum 843419 veneers required for Cortex-A53.
`bolt-worker-aarch64` runs on Neoverse V1.


ElvinaYakubova (approved):
Thanks a lot for fixing this, looks good to me!

PR #864, opened 2026-06-03
[AMD] Add `check-compiler-rt` to amdgcn bot

jhuber6:
Summary:
These are lightweight tests that check the compiler builtins and ubsan
interface right now. This should prevent us from regressing on behavior
for these. They use the same infra as the existing libc tests.


jplehr (approved):
let's see what happens

Kewen12:
Seeing error in our bot for unknown target: https://lab.llvm.org/buildbot/#/builders/10/builds/29434

jhuber6:
Strange, that should be present in https://github.com/llvm/llvm-project/blob/main/offload/cmake/caches/AMDGPULibcBot.cmake#L21

PR #865, opened 2026-06-04
Revert "[AMD] Add `check-compiler-rt` to amdgcn bot"

Kewen12:
Reverts llvm/llvm-zorg#864

unblock bot while we investigate this. https://lab.llvm.org/buildbot/#/builders/10


jplehr:
We can either revert and let Galina know (this needs a master restart) or we take the opportunity and switch the bot over to use a `ScriptedBuilder` and the cache file. This also requires a master restart and is what we want to do soon anyway.

Kewen12:
close this one as we moved with: https://github.com/llvm/llvm-zorg/pull/866

PR #868, opened 2026-06-10
Build AMDGPU profile runtime in HIP buildbot

yxsamliu:
The HIP buildbot now builds the AMDGPU profile runtime used by HIP offload PGO.


yxsamliu:
This is needed for the llvm-test-suite HIP offload PGO test: https://github.com/llvm/llvm-test-suite/pull/366

That test needs the AMDGPU profile runtime to be available in the HIP buildbot toolchain.

yxsamliu:
The failing llvm-zorg litmus test looks unrelated to this PR. It fails in buildbot checkconfig due to a syntax error in buildbot/osuosl/master/config/builders.py, introduced by 3f082442. The same error is present on current main. This PR only changes hip-build.sh.

jplehr (approved):
Long-term, I think we want to move this build recipe over to a CMake cache like we do with the other buildbots we maintain.

Objections?

yxsamliu:
No objection from me. Moving this to a CMake cache sounds cleaner and matches the other AMDGPU/offload bots.

For this PR, I would prefer to keep the change small and only enable the AMDGPU profile runtime. We can do the cache migration as a follow-up.

PR #870, opened 2026-06-11
Fix post-build step for 'llvm-clang-win-x-aarch64-release' release builder.

vvereschaka:
Use the proper properties for the devkit reboot step during SSH session.


vvereschaka:
as an example of the current step status
https://lab.llvm.org/buildbot/#/builders/1/builds/294/steps/18/logs/stdio

dyung:
Thanks for looking into this! Can we get this in and deployed to production before the last 22.x release next Tuesday by any chance so that we can have a clean build?

vvereschaka:
>Can we get this in and deployed to production before the last 22.x release next Tuesday by any chance so that we can have a clean build?

I think the buildbot configuration will be reloaded during this weekend. So, we should get it on Monday I hope.

gkistanova (approved):
LGTM.

PR #879, opened 2026-07-07
Remove flang-aarch64-debug-reverse-iteration

luporl:
It tests a very specific scenario and the worker is not being able to
keep up with the build requests.

The server that hosts it is also low on memory and disk space.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2288,22 +2288,6 @@ all += [
                         "-DCMAKE_BUILD_TYPE=Release",
                     ])},
 
-    {'name' : "flang-aarch64-debug-reverse-iteration",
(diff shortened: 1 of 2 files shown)


gkistanova (approved):
LGTM.

omjavaid (approved):
This builder was added back in the years upon community request. We cant sufficiently allocate resources for this bot anymore. This may return in case its need.
FYI @MaskRay and @DavidSpickett.

MaskRay:
> This builder was added back in the years upon community request. We cant sufficiently allocate resources for this bot anymore. This may return in case its need. FYI @MaskRay and @DavidSpickett.

Hi! Thanks for providing this service over the past few years. I just created https://github.com/llvm/llvm-project/pull/215074 so the AArch64 Linux premerge job can catch reverse iteration issues.

PR #883, opened 2026-07-13
[RISCV][lit-on-qemu] Rework implementation to just use ssh

asb:
The current implementation has a slightly awkward communication protocol including files in a shared directory. Using the approach I document here
<https://muxup.com/2026q2/bootable-qemu-image-menagerie-with-rootless-debootstrap> in terms of ssh key injection and Unix domain sockets, we instead just use ssh. Overall this should improve robustness and by virtue of being more conventional, maintainability.

The builders will be taken down temporarily to roll this out, as the appliance image configuration is updated.

### DIFF
--- a/buildbot/riscv-rise/lit-on-qemu
+++ b/buildbot/riscv-rise/lit-on-qemu
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 import os
 import pathlib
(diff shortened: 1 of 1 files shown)


lukel97 (approved):
LGTM

PR #889, opened 2026-07-15
Add CMAKE_DISABLE_PRECOMPILE_HEADERS=ON for the builders with ccache (as-builder-* workers only).

vvereschaka:
Add option to disable the precompiled headers for non-clang compilers when ccache is enabled during the build. This option avoids CMake warning:
```
CMake Warning at cmake/modules/HandleLLVMOptions.cmake:1403 (message):
  Using ccache with precompiled headers with non-Clang compilers is not
  supported and may lead to false positives.  Set
  CMAKE_DISABLE_PRECOMPILE_HEADERS to ON/OFF to silence this warning.
```

These changes get applied only to the builders hosted by as-builder-* workers.


gkistanova (approved):
LGTM.

PR #890, opened 2026-07-15
[BazelBot] Allow reaping zombie processes

pranavk:
The java server that Bazel starts can sometimes become zombie. Since PID 1 here is bazel_server_main.py which doesn't have any capability to reap zombie processes, the java server process remains alive. Subsequent Bazel builds then keep waiting for this to die and we are stuck in an infinite loop with bazelbot not doing anything useful.

shareProcessNamespace makes /pause as PID 1 which allows reaping zombie processes.

### DIFF
--- a/google-bazel-bot/bazel-fixer-bot.yaml
+++ b/google-bazel-bot/bazel-fixer-bot.yaml
@@ -8,6 +8,7 @@ spec:
   selector:
     matchLabels:
       app: bazel-fixer-bot
+  shareProcessNamespace: true
   strategy:
     type: RollingUpdate
     rollingUpdate:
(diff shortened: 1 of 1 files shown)


google-bazel-bot/bazel-fixer-bot.yaml
```diff
@@ -8,6 +8,7 @@ spec:
   selector:
     matchLabels:
       app: bazel-fixer-bot
+  shareProcessNamespace: true
```
boomanaiden154, line 12: Comment on why this is needed?

pranavk: Done

nigham (approved):

PR #893, opened 2026-07-16
Remove unused TOOLCHAIN_TARGET_SYSROOTFS CMake param for `llvm-clang-ubuntu-x-aarch64-pauthtest` builder.

vvereschaka:
The builder uses the target related rootfs there. Also it uses outdated property pointed to the root directory.


gkistanova (approved):
LGTM.

PR #894, opened 2026-07-16
[Zorg] Mark factory "always clean build" if requested for `UnifiedTreeBuilder.getCmakeExBuildFactory `.

vvereschaka:
To allow a correct buildrequest collapse for the generated builders, properly mark the factory with "always does clean build" attribute as it requested.


gkistanova (approved):
LGTM.

PR #903, opened 2026-07-29
[Buildbot][NVHPC] Add LLDB support

rglusic-nvidia:
## Summary

- enable LLDB in the NVHPC x86_64 Linux builder
- add the LLDB builder tag so the builder is discoverable under LLDB
- run `check-lldb` alongside the existing LLVM, MLIR, and Flang checks

## Testing

- `python3 -m py_compile buildbot/osuosl/master/config/builders.py buildbot/osuosl/master/config/workers.py`
- targeted static validation of the NVHPC builder, project/check lists, and worker references
- `git diff --check`


DavidSpickett:
Always happy to see more LLDB coverage. Will be running tests on the x86 host machine or are you preparing for testing GPU support, or both?

rglusic-nvidia:
This was a request from a community member for their research. CPU only right now!

buildbot/osuosl/master/config/builders.py
```diff
@@ -3885,17 +3885,18 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
                 )
         },
         {'name': "nvhpc-flang-x86_64-linux",
```
eugeneepshteyn, line 3887: AI agent flagged the following. @DavidSpickett , does this make sense to you?

`LLDB_ENABLE_PYTHON` defaults to auto-detect, and if the workers are missing Python dev headers or SWIG >= 4, cmake silently disables it and every LLDB *API* test is reported UNSUPPORTED — `check-lldb` stays green while running only the unit and shell suites. Since the upcoming Fortran tests are API tests, that would defeat the purpose of adding LLDB here. Suggest adding to the factory:

```python
extra_configure_args=[
 "-DLLDB_ENABLE_PYTHON=ON",
 "-DLLDB_ENFORCE_STRICT_TEST_REQUIREMENTS=ON",
],
```

(matching what lldb-x86_64-debian and the Linaro LLDB bots do), making sure the workers have SWIG >= 4 and the packages from `lldb/test/requirements.txt`, and doing one trial run of this configuration on a worker before merging — API tests also need ptrace to work there (relevant if the workers are containerized). Splitting `check-lldb` into `check-lldb-{unit,shell,api}` like lldb-x86_64-win does would make failure triage easier, but that's optional.

DavidSpickett: Yes that would help. The only other one I would add is `LLDB_ENABLE_LIBXML2=ON` (https://lldb.llvm.org/resources/build.html#optional-dependencies). LLDB runs a lot more smoothly with xml support enabled.

Splitting the check steps, up to you. I never felt like the results are confusing, but then again I work on lldb all the time.

eugeneepshteyn: @DavidSpickett , Ryan made suggested modifications. Do we have approval to merge once the local testing checks out?

DavidSpickett: Looks good to me.

DavidSpickett (approved):
Fine with me. This is just changing your own configuration so if you've got everything ready on your end I don't see any reason not to merge this.

I can do so if you do not have the permissions for it.

PR #905, opened 2026-08-06
[Fuchsia] Update cmake and llvm version

zeroomega:
This patch updates the cmake version to the latest on Fuchsia bot config. This patch also update the stage0 llvm version to 22.


buildbot/fuchsia/Dockerfile
petrhosek, line 2: We should also consider updating the base image to `docker.io/ubuntu:resolute`, but that should be done as a separate PR.

petrhosek (approved):

buildbot/fuchsia/Dockerfile
zeroomega: OK, I will have a separate PR to address that.

PR #907, opened 2026-08-07
[zorg] Add Arm-managed worker for clang-aarch64-quick

vrukesh:
We are migrating the clang-aarch64-quick builder from Linaro-hosted infrastructure to Arm-hosted infrastructure.

Add the new Arm-managed worker to the existing clang-aarch64-quick builder. Reusing the existing builder avoids introducing a duplicate builder during the transition and keeps the builder name unchanged.

The new worker will initially connect to the silent Buildbot master for validation. Once the existing Linaro worker has been taken offline, the Arm worker can be connected to the normal master.

The Linaro worker configuration can be removed separately, once the migration is complete.

We intend to migrate more bots in the similar fashion, once this one is successful.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -418,7 +418,7 @@ all = [
     ## AArch64 check-all
     {'name' : "clang-aarch64-quick",
     'tags'  : ["clang"],
-    'workernames' : ["linaro-clang-aarch64-quick"],
+    'workernames' : ["linaro-clang-aarch64-quick", "arm-bbot-clang-aarch64-quick"],
     'builddir': "clang-aarch64-quick",
     'factory' : ClangBuilder.getClangCMakeBuildFactory(
                     clean=False,
--- a/buildbot/osuosl/master/config/workers.py
+++ b/buildbot/osuosl/master/config/workers.py
@@ -584,4 +584,6 @@ def get_all():
                 "nvhpc-devops@nvidia.com"
             ],
         ),
+        # Arm-managed buildbot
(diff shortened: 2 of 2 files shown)


DavidSpickett:
This is subject to completing the steps in https://llvm.org/docs/HowToAddABuilder.html#steps-to-add-builder-to-llvm-buildbot of course. We (that is, Arm the company) will reach out over email to do the rest of that.

@luporl and @antmox are from Linaro. Does this transition plan make sense to you and do you have any concerns about it? I know there has been some conversation internally between Arm and Linaro, but I want to make sure we are publicly aligned too.

luporl (approved):
@DavidSpickett, this transition plan makes sense to me, as it should avoid disruptions when switching workers.
On Linaro side, the worker should be removed only when the zorg PR that removes it is picked up by the buildbot master.

DavidSpickett:
> On Linaro side, the worker should be removed only when the zorg PR that removes it is picked up by the buildbot master.

Yes.

I'll state all the steps so nothing is left unsaid:
* Linaro's worker is connected to the normal master (which is the current state).
* The details of Arm's new worker are added to the database on the llvm side.
* This PR is landed.
* The changes from this PR go live on normal and silent master.
* Arm connects their worker to the silent master.
* Once Arm's worker is confirmed to be stable, we contact Linaro and ask them to stop their worker.
* Once Linaro's worker is stopped, Arm moves its worker to normal.
* Arm is now responsible for this builder.
* Linaro's worker definitions can be removed from llvm-zorg at some point, no rush on that.

Which is a lot of back and forth, but we will do the rest in larger batches if this first one goes well.

gkistanova (approved):
LGTM.

PR #908, opened 2026-08-10
Add Intel Windows GPU machine

sarnex:
Manually tested this config on the actual machine and the builds complete in ~3 min which should allow us to keep up with traffic.

Once `libsycl` can be compiled/tested for Windows I will add support for that, but for now, just test `liboffload`.

### DIFF
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2083,7 +2083,6 @@ all += [
                         "-DRUNTIMES_spirv64-unknown-unknown_LLVM_ENABLE_RUNTIMES=compiler-rt",
                         "-DRUNTIMES_spirv64-unknown-unknown_COMPILER_RT_INCLUDE_TESTS=OFF",
                         "-DOFFLOAD_INCLUDE_TESTS=ON",
-
                     ])},
 
     {'name' : "intel-sycl-gpu-win",
--- a/buildbot/osuosl/master/config/builders.py
+++ b/buildbot/osuosl/master/config/builders.py
@@ -2082,6 +2082,8 @@ all += [
                         "-DRUNTIMES_spirv64-intel_LLVM_ENABLE_RUNTIMES=openmp",
                         "-DRUNTIMES_spirv64-unknown-unknown_LLVM_ENABLE_RUNTIMES=compiler-rt",
                         "-DRUNTIMES_spirv64-unknown-unknown_COMPILER_RT_INCLUDE_TESTS=OFF",
+                        "-DOFFLOAD_INCLUDE_TESTS=ON",
+
                     ])},
 
     {'name' : "intel-sycl-gpu-win",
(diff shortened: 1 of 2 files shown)


sarnex:
FYI @wlemkows

buildbot/osuosl/master/config/workers.py
```diff
@@ -376,6 +376,7 @@ def get_all():
 
         # SYCL GPU
         create_worker("intel-sycl-gpu-01", properties={'jobs': 192}, max_builds=1),
```
adurang, line 378: should we rename this to "gpu-lnx" or something like that?

adurang (approved):

buildbot/osuosl/master/config/workers.py
sarnex: Ideally yes, that will require updating the buildbot config on the Linux machine to use the new name, which is fine, but the problem is the time is takes for a commit in `llvm-zorg` to make it into LLVM CI production is crazy, recently it took 2 full weeks, and if we update the name on the buildbot too early, it will fail to connect from that moment so we'll lose the coverage until production is updated, and if we update it too late it will start failing to connect once production is updated until we fix it, so I just decided to not touch it.

buildbot/osuosl/master/config/builders.py
```diff
@@ -2084,6 +2084,30 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
                         "-DRUNTIMES_spirv64-unknown-unknown_COMPILER_RT_INCLUDE_TESTS=OFF",
                     ])},
 
+    {'name' : "intel-sycl-gpu-win",
```
sarnex, line 2088: this is the build config part so the name shouldnt include the gpu, but i updated the worker name to `intel-win-bmg-01` which should hopefully be more useful when it fails

buildbot/osuosl/master/config/workers.py
```diff
@@ -376,6 +376,7 @@ def get_all():
...
+        create_worker("intel-win-bmg-01", properties={'jobs': 128}, max_builds=1),
```
lplewa, line 379: is this jobs number correct - i was thinking that this machine have around ~50 threads

lplewa (approved):
@wlemkows, can you please review it too?

buildbot/osuosl/master/config/workers.py
sarnex: i think so, windows task manager shows 64 physical cores, 128 logical cores, and i've been testing locally with this number and it seems fine (3 min per job)

buildbot/osuosl/master/config/builders.py
```diff
@@ -2084,6 +2084,30 @@ def collapseRequestsDoxygen(master, builder, req1, req2):
...
+    'tags'  : ["sycl"],
+    'collapseRequests': False,
+    'workernames' : ["intel-win-bmg-01"],
+    'builddir': "intel-sycl-gpu-win",
+    'factory' : UnifiedTreeBuilder.getCmakeWithNinjaBuildFactory(
+                    clean=True,
+                    checks=['check-offload'],
```
wlemkows, line 2095: Do we need OFFLOAD_INCLUDE_TESTS here?

sarnex: it seems to be working without it, but i added it in https://github.com/llvm/llvm-zorg/pull/908/commits/158d06ab40497304ab4422f3262b2e66816d9d6c to be safe, thanks

sarnex:
merging now as it takes forever for changes to make it into production, if @wlemkows has any more feedback i will address it in a follow-up PR, thanks!

PR #916, opened 2026-08-20
[zorg] Add Arm-managed workers for AArch64 Flang, LLDB and GlobalISel

vrukesh:
We are migrating the AArch64 Flang, LLDB and GlobalISel builders from Linaro-hosted infrastructure to Arm-hosted infrastructure.

Add the new Arm-managed workers to the existing AArch64 Flang, LLDB and GlobalISel builders. Reusing the existing builders avoids introducing duplicate builders during the transition and keeps the builder names unchanged.

The new workers will initially connect to the silent Buildbot master for validation. Once the existing Linaro workers have been taken offline, the Arm workers can be connected to the normal master.

The Linaro worker configuration can be removed separately, once the migration is complete.


gkistanova (approved):
LGTM.

PR #921, opened 2026-08-26
[CI] Disable konnectivity-agent autoscaling

boomanaiden154:
This should allows us to use the kubernetes executor mode which makes
maintenance much easier, especially for libc++. Details on why are
mostly in the comment added in the PR.

### DIFF
--- a/premerge/premerge_resources/main.tf
+++ b/premerge/premerge_resources/main.tf
@@ -32,7 +32,7 @@ resource "kubernetes_manifest" "konnectivity_autoscaler_replica_count_override" 
       namespace = "kube-system"
     }
     spec = {
-      repliaes = 0
+      replicas = 0
     }
   }
 
@@ -41,7 +41,7 @@ resource "kubernetes_manifest" "konnectivity_autoscaler_replica_count_override" 
   }
 }
 
-resource "kubernetes_manifest" "konnectivity_agent_relica_count_override" {
+resource "kubernetes_manifest" "konnectivity_agent_replica_count_override" {
   manifest = {
     apiVersion = "apps/v1"
     kind       = "Deployment"
--- a/premerge/premerge_resources/main.tf
+++ b/premerge/premerge_resources/main.tf
@@ -32,7 +32,7 @@ resource "kubernetes_manifest" "konnectivity_autoscaler_replica_count_override" 
       namespace = "kube-system"
     }
     spec = {
-      replices = 0
+      repliaes = 0
     }
   }
 
@@ -50,7 +50,7 @@ resource "kubernetes_manifest" "konnectivity_agent_relica_count_override" {
       namespace = "kube-system"
     }
     spec = {
-      replices = 10
+      replicas = 10
     }
   }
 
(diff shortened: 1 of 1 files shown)


premerge/premerge_resources/main.tf
```diff
@@ -11,6 +11,54 @@ terraform {
...
+      replices = 0
+    }
+  }
+
+  field_manager {
+    force_conflicts = true
+  }
+}
+
+resource "kubernetes_manifest" "konnectivity_agent_relica_count_override" {
```
cmtice: typo: You dropped the 'p' from 'replica'.

premerge/premerge_resources/main.tf
```diff
@@ -11,6 +11,54 @@ terraform {
...
+resource "kubernetes_manifest" "konnectivity_autoscaler_replica_count_override" {
+  manifest = {
+    apiVersion = "apps/v1"
+    kind       = "Deployment"
+    metadata = {
+      name      = "konnectivity-agent-autoscaler"
+      namespace = "kube-system"
+    }
+    spec = {
+      replices = 0
```
cmtice: Is 'replices' really the right spelling? Not "replicas'? (you did spell it the same way in both records...)

premerge/premerge_resources/main.tf
```diff
@@ -11,6 +11,54 @@ terraform {
...
 
+# konnectivity-agent pods proxy some kubernetes API request to the backplane,
+# including kubectl exec calls. By default, GKE has a horizontal autoscaler
+# setup to provision a number of konnectivity-agent pods based purely on the
+# total node count of the cluster. A konnectivity-agent pod getting turned off
+# during a downscale will drop all of its running kubectl exec connections. This
+# means that runners running in kubernetes mode will randomly fail. We prevent
+# this by disabling autoscaling and setting a fixed count of konnectivity-agent
+# pods that is similar to what the horizontal autoscaler would use at max size.
+#
```
cmtice, line 22: General question: This means we're turning off autoscaling and going back to the original (old system) way of having just a fixed number of runners?

boomanaiden154: No. The runner pods/nodes will still autoscale like normal. This only impacts the `konnectivity-agent` pods which run on the service cluster which is on all the time. Instead of dynamically scaling the pod count (which get scheduled in otherwise unused space in the service cluster), we just set it to a fixed count.

premerge/premerge_resources/main.tf
cmtice: This typo is still here.

premerge/premerge_resources/main.tf
```diff
@@ -11,6 +11,54 @@ terraform {
...
+resource "kubernetes_manifest" "konnectivity_autoscaler_replica_count_override" {
+  manifest = {
+    apiVersion = "apps/v1"
+    kind       = "Deployment"
+    metadata = {
+      name      = "konnectivity-agent-autoscaler"
+      namespace = "kube-system"
+    }
+    spec = {
+      repliaes = 0
```
cmtice: I don't think "repliaes" is quite correct, either.

premerge/premerge_resources/main.tf
boomanaiden154: Fixed for real this time.

cmtice (approved):

PR #926, opened 2026-08-27
[Buildbot][z/OS] Disable threads for z/OS builder and reduce jobs for worker

GaminRick7:
This PR disables threads for the z/OS builder `llvm-s390x-zos` and reduces the jobs from 4 to 1 for the `zos-s390x-1` worker.


gkistanova (approved):
LGTM.

PR #929, opened 2026-09-02
Add a new NV CUDA Offload "nvhpc-offload-nvgpu-x86_64-linux" builder.

vvereschaka:
NVIDIA GPU Offload builder (flang/openmp) on Linux/x86-64 build host.

PR #934, opened 2026-09-05
The builder configuration adjustment for 'nvhpc-offload-nvgpu-x86_64-linux'.

vvereschaka:
Explicitly specify the lists of the enabled projects (`enable_projects`) and the enabled runtimes (`enable_runtimes`) for the builder. We should not use `auto` for them as we use the `depends_on_projects` parameter to pass all dependencies for all stages.

Ref: #929


vvereschaka:
@gkistanova @ivanradanov @vzakhari JFR/FYI

vvereschaka:
Just in case
https://lab.llvm.org/staging/#/builders/239 on staging, to check the status.

PR #939, opened 2026-09-10
[Arm] Do not use lld to link the bulk of things on armv8-lld-2stage

DavidSpickett:
Since the clang unit tests were combined into a single binary, we are hitting RAM limits on this 32-bit bot
(https://github.com/llvm/llvm-project/issues/172659).

We could mess about with options but that's going to take a lot of time to prove, and unlikely to work if we are simply out of RAM. What I do know is that we have armv7 (32-bit) bots building and testing clang just fine with the system ld.

So we're going to use that as the main linker instead. lld will be built and tested and used to link llvm-test-suite programs.


DavidSpickett:
@vrukesh fyi

DavidSpickett:
Since this is our bot (will be once Linaro switches theirs off), and it's already running on silent, I'm going to merge this now without approval.

PR #941, opened 2026-09-10
[BazelBot] Sync with upstream buildkite config

pranavk:
Upstream buildkite config stopped using query a while back. This PR syncs it to that so Bazel fixer bot builds the same thing as upstream.


boomanaiden154 (approved):
LGTM.

We should probably figure out a way to consolidate these scripts at some point. I don't think the `-test*` ones are used.

rupprecht (approved):

pranavk:
Only bazel-build-ci is used right now. Yeah, we should just have one script with options.

PR #948, opened 2026-09-15
[Buildbot][z/OS] Change diff exe path for z/OS builder

GaminRick7:
This PR changes `DIFF_EXE` to use an upgraded diff from GNU Diffutils for z/OS builder llvm-s390x-zos.


abhina-sree (approved):
LGTM

PR #954, opened 2026-09-17
[zorg] Add new Linaro-hosted WoA workers

ceseo:
Adding eight new Windows on Arm Linaro-hosted workers. These will be used to run:

 * clang-arm64-windows-msvc
 * clang-arm64-windows-msvc-2stage
 * lldb-arm64-windows
 * flang-arm64-windows-msvc
 * flang-arm64-windows-msvc-testsuite


omjavaid:
Hi Carlos I dont think we need these workers right now. We can use the existing worker names for testing on staging. Final migration might take a while so may be we can hold these changes untill everything has been fully tested.

